PR #5, opened 2026-08-18
claude: effort selector via extended thinking + 5-series models

eternally-black:
The `claude` route is the only one of the three without an effort selector, and its built-in catalog stops at 4.5. Both are fixable and I verified the assumption behind the first one against the live endpoint.

## Extended thinking works on the subscription endpoint

`resolveModel` carried this note:

> No reasoning metadata: the subscription endpoint's thinking support is not exercised, so effort requests reject as unsupported.

It is supported. With exactly the beta flags this provider already sends (`claude-code-20250219,oauth-2025-04-20,interleaved-thinking-2025-05-14,context-management-2025-06-27`) and the Claude Code identity headers, a request with `thinking: {"type": "enabled", "budget_tokens": 4096}` returns HTTP 200 and streams `thinking` / `thinking_delta` / `signature_delta` blocks before the text. Checked on a Max subscription today, against `claude-opus-4-5` and the 5-series.

So the route now advertises `low`/`medium`/`high` (default `medium`) and translates the effort into a thinking budget on the wire — the Anthropic API has no `effort` field, it has a token budget. `max_tokens` covers the thinking budget too, so it is lifted above `budget + 4096` when the configured ceiling is tighter; otherwise the configured value stands.

## 5-series in the default catalog

`claude-opus-5`, `claude-sonnet-5` and `claude-fable-5` all answer 200 on the subscription endpoint. Unlike codex and grok, the claude route has no models endpoint to discover from, so `DEFAULT_MODELS.claude` is what users see until they override it in config — worth keeping current. The 4.5 entries stay where they are.

## Tests

Four assertions added to `test/models.spec.ts`: the effort metadata, the thinking block on the wire, absence of the block when no effort is set, and the `max_tokens` lift. Full suite: 75/75.

Note on running the suite locally: `devDependencies` point at `link:/Users/v1ki/...` paths, so a fresh clone cannot install them. I ran the tests against the published `@deepseek-ai/*` packages from a dsh profile. Might be worth switching those links to versioned deps so outside contributors can run `pnpm test` — happy to send that separately if useful.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -116,6 +116,9 @@ const DEFAULT_MODELS: Record<ProviderId, ModelEntry[]> = {
     { id: 'gpt-5.1', name: 'GPT-5.1' },
   ],
   claude: [
+    { id: 'claude-opus-5', name: 'Claude Opus 5' },
+    { id: 'claude-sonnet-5', name: 'Claude Sonnet 5' },
+    { id: 'claude-fable-5', name: 'Claude Fable 5' },
     { id: 'claude-opus-4-5', name: 'Claude Opus 4.5', maxTokens: 64_000 },
     { id: 'claude-sonnet-4-5', name: 'Claude Sonnet 4.5' },
     { id: 'claude-haiku-4-5', name: 'Claude Haiku 4.5' },
--- a/src/providers/claude.ts
+++ b/src/providers/claude.ts
@@ -4,7 +4,7 @@
  * the Anthropic Messages API with the Claude Code identity headers.
  */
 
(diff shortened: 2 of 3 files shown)


V1ki:
感谢贡献！这个 PR 主要增加静态 Claude 5 系列模型列表以及 extended thinking / effort 支持；这些功能已经由 #2 覆盖。#2 会动态发现可用模型，并根据模型实时返回的 capabilities 动态维护 thinking 类型和 effort levels，因此继续合并这里的静态模型列表和独立 effort 映射会形成重复实现，并与动态发现结果产生冲突。基于以上原因先关闭此 PR。测试思路仍然很有价值，后续可以考虑将相关测试适配到 #2 的动态实现。

PR #6, opened 2026-08-19
fix: classify codex usage windows by reported duration, not slot （修正周限额被显示为 5 小时限额问题）

shpirt:
## Problem

The wham/usage backend has been observed to report the weekly lane as `primary_window` with no `secondary_window`. `fetchCodexUsage` hardcoded `primary_window` to the 5-hour session lane and `secondary_window` to the weekly lane, so a weekly quota was displayed under the "5-hour window" label instead of "Weekly".

## Change

Classify each wham/usage window by its reported `limit_window_seconds`:

- ~18000s (5 hours) → `session`
- ~604800s (7 days) → `weekly`
- anything else → `other`

The slot-based mapping is kept only as a fallback when the duration is absent, so unknown window shapes no longer get mislabeled.

## Tests

- weekly-only primary window is classified as `weekly`
- unrecognized durations map to `other`
- missing duration falls back to slot-based kind

### DIFF
--- a/src/providers/codex.ts
+++ b/src/providers/codex.ts
@@ -239,14 +239,39 @@ export const CODEX_USAGE_URL = 'https://chatgpt.com/backend-api/wham/usage'
 /** One `rate_limit.*_window` object of the wham/usage payload (subset). */
 interface CodexUsageWindow {
   used_percent?: number
+  /** Window duration in seconds (18000 = 5 hours, 604800 = 7 days). */
+  limit_window_seconds?: number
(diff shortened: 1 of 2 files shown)


V1ki:
Thanks @shpirt — merged.

Confirmed on a Pro account: `wham/usage` reported the weekly lane as `primary_window` (`limit_window_seconds: 604800`) with no `secondary_window`, so the old slot mapping labeled it as the 5-hour window. Classifying by duration is the right fix.

PR #10, opened 2026-08-19
feat: OAuth fallback when Claude Code CLI is unavailable

### LINKED ISSUE #9: Claude: OAuth fallback when Claude Code CLI is not installed
## Context

PR #2 replaced Claude's interactive OAuth login with direct credential import from Claude Code's own store (macOS Keychain / `~/.claude/.credentials.json`). This is the preferred path — it avoids a second login, enables token synchronization, and eliminates the account-risk concerns of running an independent OAuth session.

However, as @V1ki [noted in review](https://github.com/V1ki/dsh-plugin-subscriptions/pull/2#issuecomment-5340393498), this makes Claude Code a hard requirement. Environments without the CLI installed — headless hosts, containers, remote servers, clean machines — currently have no way to connect a Claude subscription.

## Proposal

Add an OAuth fallback path for Claude login:

1. **Try credential import first** — if a valid Claude Code credential store exists, import instantly (current behavior).
2. **Fall back to interactive OAuth** — if no CLI/Keychain/credential file is available, start the browser-based OAuth flow (the code for `claudeFlow`, `exchangeClaudeCode()`, etc. is still present in the codebase).
3. **Shared refresh/storage** — both paths produce a `ClaudeSession`; refresh and persistence work identically regardless of how the session was obtained.

```ts
async login(provider: ProviderId): Promise<{ authorizeUrl: string }> {
  if (provider === 'claude') {
    const imported = readClaudeCodeCredentials()
    if (imported !== undefined) {
      await this.persist('claude', imported)
      this.onAuthChanged('claude')
      return { authorizeUrl: '' }
    }
    // No Claude Code installation — fall back to browser flow
    const attempt = await this.flows.start('claude', claudeFlow)
    void this.complete('claude', attempt)
    return { authorizeUrl: attempt.authorizeUrl }
  }
  // ...
}
```

## Test coverage needed

- Credentials present → instant import, no OAuth attempt
- Credentials absent → returns a non-empty authorization URL, completes via code exchange
- Imported credentials malformed/expired → falls back to OAuth
- Refresh works for sessions produced by either path

delef:
## Summary

Add OAuth fallback for Claude login when Claude Code CLI is not installed, and fix the OAuth authorize flow to work with the current endpoint.

### What changed

1. **OAuth fallback** — `login('claude')` tries credential import first; if no credentials found, falls back to interactive browser OAuth instead of throwing an error.

2. **Fixed OAuth authorize URL** — matched Claude Code 2.1.235 exactly:
   - Endpoint: `https://claude.ai/oauth/authorize` (was `claude.com/cai/oauth/authorize`)
   - `code=true` parameter restored
   - `redirect_uri` uses local callback `http://localhost:<port>/callback` (was hardcoded `platform.claude.com`)
   - Full scope with `org:create_api_key` restored

3. **Fixed 32-byte state** — the authorize endpoint rejects 16-byte state values with "Invalid request format". Changed `OAuthFlowManager` to use `randomToken(32)` matching Claude Code's entropy.

4. **`_readClaudeCreds` test injection** — `SubscriptionsAuthController` accepts an optional credential reader for testing both paths without mocking modules.

### Files changed

- `src/providers/claude.ts` — authorize URL, scope, redirect_uri, client_id
- `src/auth/oauth-flow.ts` — state entropy 16→32 bytes
- `src/index.ts` — fallback logic + `_readClaudeCreds` injection
- `test/login.spec.ts` — 7 tests (5 unit + 2 integration via RPC mount)

Closes #9

## Test plan

- [x] With Claude Code installed: Connect → instant import, no browser opens
- [x] Without Claude Code credentials: Connect → opens claude.ai OAuth page
- [x] OAuth flow completes → "Login successful" page, session stored
- [x] OAuth URL parameters match Claude Code 2.1.235 exactly
- [x] 32-byte state accepted by endpoint (16-byte was rejected)
- [x] Unit tests: valid/invalid/malformed/bare credentials parsing
- [x] Integration tests: import path (authorizeUrl='') and fallback path (authorizeUrl starts with claude.ai)
- [x] Credential import + OAuth session both refresh correctly


delef:
@V1ki Tested both paths end-to-end on macOS with Claude Code 2.1.235.

### 1. Credential import path (Claude Code installed)

Settings → Claude → Connect → session imported instantly from macOS Keychain, no browser opens, `authorizeUrl: ''`. Models load, thinking/effort work.

### 2. OAuth fallback path (no credentials)

Hid credentials to simulate a clean environment:
```bash
# Backup + remove Keychain entry
security find-generic-password -s "Claude Code-credentials" -w > /tmp/backup.txt
security delete-generic-password -s "Claude Code-credentials"
# Rename credentials file
mv ~/.claude/.credentials.json ~/.claude/.credentials.json.bak
```

Restarted dsh → Settings → Claude → Connect → browser opens `claude.ai/oauth/authorize`, approve → "Login successful. You can close this tab and return to DeepSeek Harness." Session stored, models load.

### 3. OAuth URL debugging

We iterated through several authorize URL issues before landing on the working combination:

| Parameter | Old (broken) | Fixed | Source |
|-----------|-------------|-------|--------|
| Endpoint | `claude.com/cai/oauth/authorize` | `claude.ai/oauth/authorize` | Captured from `claude auth login` |
| `redirect_uri` | `https://platform.claude.com/oauth/code/callback` | `http://localhost:<port>/callback` | Local callback server |
| `state` entropy | 16 bytes (22 chars) | 32 bytes (43 chars) | **Root cause of "Invalid request format"** |
| `code` | removed then restored | `true` | Matches Claude Code |
| `org:create_api_key` in scope | removed then restored | present | Matches Claude Code |

The 16→32 byte state change was the critical fix — the endpoint rejects short state values with "Invalid request format". Verified by generating identical URLs differing only in state length: 16-byte fails, 32-byte succeeds.

### 4. Tests

7/7 pass (`node --test --test-force-exit lib-test/test/login.spec.js`):
- 5 unit tests for `readClaudeCodeCredentials` (valid, malformed, incomplete, bare fields, absent)
- 2 integration tests via RPC mount with `_readClaudeCreds` injection:
  - credentials present → `authorizeUrl === ''`
  - credentials absent → `authorizeUrl.startsWith('https://claude.ai/oauth/authorize')`

V1ki:
Thanks for the PR and for tracking the latest Claude Code OAuth flow. I reviewed the current head and found a few issues that need to be addressed before this can be merged:

1. **The test-only config field breaks TypeScript compilation**

   `Config` now includes `_readClaudeCreds?: () => ClaudeSession | undefined`, but the exported Schemastery `Config = z.object(...)` does not define that field. As a result, `tsc` fails in `src/index.ts` because the schema shape no longer matches the `Config` interface.

   Since functions are not serializable plugin configuration, it would be better not to expose this injection hook through the public config. Please inject the dependency through an internal controller/factory or another test-only mechanism instead.

2. **The new OAuth fallback test asserts the old authorize endpoint**

   The implementation now correctly uses `https://claude.ai/oauth/authorize`, but `test/login.spec.ts` still expects `https://claude.com/cai/oauth/authorize`. Running the new login tests in an isolated `DSH_HOME` gives 6 passes and 1 failure on this assertion. Please update the test, preferably using `CLAUDE_AUTHORIZE_URL` and also verifying the important URL parameters.

3. **The tests may write fake credentials into a developer's real DSH home**

   The test currently uses:

   `process.env.DSH_HOME = process.env.DSH_HOME ?? mkdtempSync(...)`

   If `DSH_HOME` is already set, the integration test persists `FAKE_SESSION` into the real `$DSH_HOME/plugins/subscriptions/auth.json`. Please always force an isolated temporary DSH home before importing the plugin, and clean it up afterward (or inject an explicit store path).

Additionally, this PR is currently conflicting with the latest `main`, notably around the updated auth RPC/speed-controller wiring in `src/index.ts`.

The project now supports the latest **DSH rc.08** on `main`. Please rebase this PR onto the latest `main`, resolve the conflicts, and make the fixes/test updates against the current rc.08-compatible codebase.

Once these points are addressed, I’ll be happy to take another look.

delef:
@V1ki All four points are fixed and the branch is rebased onto `0f9c330` (v0.5.0, rc.08) — the PR reads `MERGEABLE` / `CLEAN` again.

Apologies for the state it arrived in. That push went out late at night and I "verified" it with `tsdown` and `tsc --noCheck` — neither of which type-checks — and never re-ran the tests after changing the very constant one of them asserts. The gate is `pnpm build && pnpm test` from now on.

**1. Test-only config field.** `_readClaudeCreds` is gone from `Config` entirely — no trace left in `src/`, `test/` or `lib/`. The credential reader is now a constructor parameter on `SubscriptionsAuthController` carrying the production default, so the plugin always passes the real one and nothing about it reaches the schema. `stripInternal` is enabled as well, so the class stays out of the published `lib/index.d.ts`.

**2. Authorize endpoint in the test.** The test now checks the URL against `CLAUDE_AUTHORIZE_URL` and verifies the parameters: `code`, `client_id`, `response_type`, `scope`, `code_challenge_method`, the loopback `redirect_uri` with its ephemeral port, and the state entropy. One extra test pins those constants to literals — without it every assertion would compare a value against itself, and the exact regression you caught would still pass.

I also corrected the commit message, which was wrong: `claude.com/cai/oauth/authorize` is alive and 307-redirects to `claude.ai/oauth/authorize` with the query preserved. Sending `claude.ai` directly is what Claude Code does and drops a hop; it is not a fix for a breakage, and the commit no longer claims otherwise.

On the state change, for the record: with every other parameter held identical — same host, same port, same scope, same PKCE method — the endpoint answered `Invalid request format` to the 22-character state this flow used to send and accepted the 43-character one.

**3. Fake credentials in a real DSH home.** Each test now scopes its own `mkdtemp` home to the call rather than the process, restores the previous value afterwards, removes the directory, and asserts that `authFilePath()` resolves inside it. Verified by running the suite with `DSH_HOME` already set: nothing is written there. As a side effect the suite also stopped hijacking `rpc.spec`'s home in the aggregate run.

**4. Rebase.** Done, with the auth RPC / speed-controller wiring intact.

While reworking it I found a bug in my own change and fixed it: the import path did not cancel an OAuth attempt left in flight. Before this PR `pending('claude')` was always empty, so it could not happen. Now a user who clicks Connect without credentials, logs in through the CLI and clicks again would otherwise leave the callback listener bound, `busy` stuck on in the Settings card, and the still-open browser tab able to overwrite the session just imported. There is a test for it.

Suite: 106 tests, 100 pass, 0 fail, 6 skipped — the credential-file tests now skip on machines whose Keychain answers the read, instead of reporting a pass with their assertions silently disabled. Six mutations were checked to fail the suite: the authorize host, a dropped scope, the hosted `redirect_uri`, a 16-byte state, the removed cancel, and the removed fallback branch.

Two pre-existing issues I deliberately left alone — both upstream, both newly reachable now that claude enters the OAuth path:

- `OAuthAttempt.manual` does not parse the `code#state` form the console callback displays, so pasting it sends the whole string as `code`.
- `OAuthFlowManager.start` checks `attempts.has(provider)` before `await listen(...)`, so two near-simultaneous logins both bind a server and the first is orphaned until its timeout.

Happy to open separate issues for those if you'd like them tracked.

V1ki:
Thanks for addressing the previous review points and rebasing onto the rc.08-compatible main. I re-reviewed the latest head: the config/schema issue, stale URL assertion, test-home isolation, and merge conflicts are all fixed.

I found one remaining race that should be addressed before merging:

### Stale OAuth completion can overwrite a newer Claude credential import

In `src/index.ts`, the import path calls:

`this.flows.pending('claude')?.cancel()`

This only works while the OAuth attempt is still waiting for its callback. Once the callback delivers a code, `OAuthFlowManager` removes the attempt from its pending map immediately, while `complete()` continues the token exchange and later persists the returned session.

That leaves this sequence possible:

1. Claude OAuth callback delivers its code.
2. The attempt disappears from `pending('claude')`, but token exchange is still running.
3. The user logs in through Claude Code and clicks Connect again.
4. The new Claude Code credentials are imported and persisted; the cancel call is a no-op.
5. The older OAuth exchange finishes and `complete()` overwrites the newly imported session.

I reproduced this deterministically without calling the real Claude service by mocking `globalThis.fetch`: the token endpoint response was held on a promise, the local callback was invoked with the correct code/state, Claude Code credentials were then imported, and finally the old token response was released. The observed values were:

`after callback: { busy: false, pending: false }`
`after import: imported-cli`
`after old exchange: old-oauth`

The same lifecycle issue means a logout during this exchange window can also be undone when the old OAuth completion persists its session.

Please add attempt generation/identity validation (or cancellation propagated through an `AbortSignal`) and verify immediately before persistence that the completion is still current. Suggested regression tests:

- callback delivered + exchange delayed + newer credential import → old exchange must not overwrite the imported session;
- callback delivered + exchange delayed + logout → old exchange must not restore the session.

Apart from this race, the previous requested changes look resolved and the new OAuth/login tests pass in my isolated run.

delef:
Fixed in bb7c6b4 and 2538ac7.

The attempt leaving `pending()` at callback time was exactly it. I reproduced your sequence as two regression tests before touching the fix and got the same values you did — `old-oauth` winning both the import and the logout.

Ownership is now a per-provider claim counter instead of something read off the flow manager. Starting a login, importing Claude Code credentials, `cancel` and `logout` each take the claim; a completion persists its session only while the claim it started with is still current. The claim is taken after `flows.start` resolves, so a rejected second login (one attempt per provider) does not supersede the attempt already running.

Chasing it surfaced a second one underneath: `saveSession`/`deleteSession` are unlocked read-modify-write over a single file, so two providers refreshing tokens near the same moment lose one of the two sessions. Store writes now join a per-path chain, joined synchronously so call order is write order. Separate commit, with its own spec.

Both regression tests you suggested are in `test/login.spec.ts`: `globalThis.fetch` mocked with the token response held on a promise, the callback delivered to the real loopback server, the racing actor run, then the response released. `settled()` on the controller is new and test-only — a login's exchange outlives the `login()` call that started it, so asserting on what it stored would otherwise mean guessing at a timeout.

Each guard was checked by removing it and watching the matching tests go red, not by assertion. 112 tests, 106 pass, 6 skipped — a Claude Code Keychain entry shadows the credentials-file fixtures on this machine.

V1ki:
Thanks for the thorough follow-up — the claim counter, the serialized store writes, the isolated login tests, and the rebase onto rc.08 all look good.

One small leftover before this can be merged:

**A superseded OAuth exchange can still write `lastError` after a newer login has already succeeded.**

In `SubscriptionsAuthController.complete()`, the claim is checked only on the success path, immediately before `persist()`. The `catch` block still records the failure unconditionally (except for `'login cancelled'`):

```ts
if (this.claims.get(provider) !== claim) return
await this.persist(provider, session)
this.lastError.delete(provider)
this.onAuthChanged(provider)
} catch (error) {
  if (!(error instanceof Error && error.message === 'login cancelled')) {
    this.lastError.set(provider, errorChain(error))
  }
}
```

That leaves this sequence:

1. Claude OAuth callback is delivered and token exchange is still in flight.
2. The user imports Claude Code credentials (or otherwise takes a newer claim).
3. The imported session is persisted and `lastError` is cleared.
4. The old exchange then fails on the network.
5. `catch` writes `lastError` even though the current claim no longer owns the provider.

The Settings card then shows an error on a provider that is already logged in.

Please also check the claim in `catch` (same `claims.get(provider) !== claim` guard) so a stale failure cannot surface after a newer import/login/logout. A regression next to the existing delayed-exchange tests would be ideal: callback delivered + exchange rejected + newer import → status has no `detail`.

Once that is in, I’ll merge.

delef:
@V1ki done

V1ki:
Merged, thank you.

This was a careful piece of work — the OAuth fallback, the authorize-flow alignment with Claude Code, and the follow-up on the completion/claim races and tests. Appreciate you sticking with the review rounds.

PR #23, opened 2026-08-22
fix(claude): keep replayed tool calls and tool_result ordering valid on the wire

### LINKED ISSUE #22: Claude route 400s permanently once a background subagent settles (`tool_use` blocks can only be in `assistant` messages)
## Symptom

Turns through the `claude` provider fail with HTTP 400 once a background subagent has settled in the session:

```
claude API error (HTTP 400): {"type":"error","error":{"type":"invalid_request_error",
"message":"messages.52: `tool_use` blocks can only be in `assistant` messages"},
"request_id":"req_011CeFgHmAGnSaUyKwBCDbGU"}
```

The failure is sticky: the offending message stays in history, so every subsequent turn in that session fails the same way. The `codex` and `grok` routes are unaffected in the same session.

## Root cause in v0.5.0

When a background subagent settles, DSH splices that subagent's **closing assistant message** into the parent conversation as a **`user`-role notice** (`source.kind: "subagent-settled"`, `form: "notice"`), carrying the subagent's blocks verbatim — including the `tool-call` blocks it was holding when it died.

`toAnthropicMessages()` maps blocks to wire form by block type alone and never consults `message.role` (`src/translate/anthropic.ts`):

```ts
for (const block of message.content) {
  switch (block.type) {
    case 'tool-call':
      blocks.push({
        type: 'tool_use',
        id: String(block.id),
        name: block.name,
        input: parseToolInput(block.arguments),
      })
      break
```

So the replayed tool call reaches the wire as a `tool_use` block inside a `user` message, which the Messages API rejects.

Re-roling the notice to `assistant` would not help: the subagent died mid-flight, so those `tool_use` ids have no matching `tool_result` and the request would be rejected for the dangling call instead.

`toResponsesInput()` is immune because the Responses API models a call as a top-level `function_call` input item rather than a block inside a role-tagged message, so the same history translates legally for `codex` and `grok`:

```ts
case 'tool-call':
  flushMessage()
  input.push({ type: 'function_call', call_id: String(block.id), name: block.name, arguments: block.arguments })
  break
```

The harness type permits this: `Message.role` is `'system' | 'user' | 'assistant'` and `content` is an unrestricted `ContentBlock[]`, so a `user` message carrying `reasoning` and `tool-call` blocks is well-formed input that only the Anthropic route constrains.

## Evidence

Session `acede550-e4b5-4c4a-848a-c1441de36e22`, message 70 of the durable log:

```
type:   user/message
role:   user
source: { kind: "subagent-settled", form: "notice",
          summary: "Background subagent 7f21c45a-... failed before it finished.",
          senderSessionId: "7f21c45a-..." }
content: text, text, reasoning, tool-call, tool-call
```

The two `tool-call` blocks are `bash` calls (`toolu_01MG4h1aCn3qyim5LzNFSBz6`, `toolu_01BsBzcajiRawyeUERTqZbFE`) with no `tool-result` anywhere later in the log.

Replaying that session's 72 harness messages through `toAnthropicMessages()` yields 53 Anthropic messages, and index **52** is exactly that notice — matching the `messages.52` the API named:

```
index 52  role: user  blocks: tool_result, text, text, tool_use, tool_use, text
```

Replaying all 20 sessions under `~/.dsh/sessions` (537 Anthropic messages) against the full structural contract — role/block agreement, `tool_use` ↔ `tool_result` pairing and adjacency, `tool_result` leading its user message, non-empty content, alternating roles — flags exactly one session and exactly one index:

```
sessions replayed: 20   anthropic messages produced: 537   clean: 19

FAIL session-acede550-…  (53 messages)
  messages.52: tool_use-outside-assistant       user
  messages.52: tool_use-outside-assistant       user
  messages.52: dangling-tool_use-at-end         toolu_01MG4h1aCn3qyim5LzNFSBz6,toolu_01BsBzcajiRawyeUERTqZbFE
```

The third line is why re-roling is not the fix: the calls are unanswered as well as misplaced.

### API-level confirmation

A minimal synthetic request through the plugin's own headers (`claude-haiku-4-5-20251001`, `max_tokens: 16`), differing only in how the notice's tool call is rendered:

| Notice renders the tool call as | Result |
|---|---|
| `tool_use` block | **HTTP 400** — ``messages.2: `tool_use` blocks can only be in `assistant` messages`` |
| `text` block | **HTTP 200** |

So the rejection is reproducible in isolation, and the proposed rendering is accepted by the live endpoint rather than merely by local reasoning.

## Impact

A settled background subagent permanently breaks the `claude` route for that session. There is no in-session recovery — every retry re-sends the same history — so the session has to be abandoned. Sessions that use background subagents are affected disproportionately, and the failure surfaces only after the subagent settles, well away from the code that caused it.

## Proposed resolution

Branch on `message.role` before emitting `tool_use`. A tool call outside an assistant message is replayed narrative with no result to pair with, so render it as descriptive text rather than a call the API will reject:

```ts
case 'tool-call':
  blocks.push(role === 'assistant'
    ? { type: 'tool_use', id: String(block.id), name: block.name, input: parseToolInput(block.arguments) }
    : { type: 'text', text: `[tool call ${block.name}: ${block.arguments}]` })
  break
```

This keeps the notice's information — what the subagent was doing when it died — instead of dropping the blocks, and the request becomes valid:

```
index 52  role: user  blocks: tool_result, text, text, text, text, text
```

With the branch in place the same 20-session sweep is clean:

```
sessions replayed: 20   anthropic messages produced: 537   clean: 20

no structural violations
```

I have this working locally with a regression test and can open a PR.

### Adjacent risk, not covered here

`tool_result` blocks must lead a user message. Today notices target `next-turn` and land after the turn's tool results, so the merged message keeps `tool_result` first (confirmed above). If a future injection point splices a notice ahead of a tool result in the same merged user message, the request would fail for block ordering instead. Worth keeping in mind, but there is no reproduction for it.

## Environment

- `dsh-plugin-subscriptions` 0.5.0
- Claude Pro/Max subscription, `claude-opus-5`
- DSH web profile, background subagents enabled

delef:
Fixes #22.

## The defect

When a background subagent settles, DSH splices its closing assistant message into the parent conversation as a **`user`-role notice** (`source.kind: "subagent-settled"`, `form: "notice"`), carrying that message's blocks verbatim — including the `tool-call` blocks it was holding when it died.

`toAnthropicMessages()` mapped blocks to wire form by block type alone, never consulting `message.role`, so those replayed calls reached the wire as `tool_use` blocks inside a `user` message:

```
claude API error (HTTP 400): messages.52: `tool_use` blocks can only be in `assistant` messages
```

The failure is sticky — the notice stays in history, so every later turn in that session fails identically and the session has to be abandoned. `codex` and `grok` are unaffected because the Responses API models a call as a top-level `function_call` input item rather than a block inside a role-tagged message.

## The change

Branch on the message role before emitting `tool_use`. A tool call outside an assistant message is replayed narrative, and — because the subagent died mid-flight — its ids have no matching `tool_result`, so re-roling it to `assistant` would only trade one rejection for another. It rides as descriptive text instead, which keeps what the notice is *for* (what the subagent was doing when it died) rather than dropping the blocks:

```
[tool call bash: {"command": "grep -rn \"UI hint\" --include=*.md ."}]
```

## Verification

**Real session replay.** Session `acede550-…` is the reported failure. Replaying its 72 harness messages produces 53 Anthropic messages, and index **52** is the notice — matching the `messages.52` the API named. Before: `tool_result, text, text, tool_use, tool_use, text`. After: `tool_result, text, text, text, text, text`.

**Full structural sweep.** All 20 sessions on disk (537 Anthropic messages) checked against the contract — role/block agreement, `tool_use` ↔ `tool_result` pairing and adjacency, `tool_result` leading its user message, non-empty content, alternating roles, first-message-is-user:

| | clean | violations |
|---|---|---|
| before | 19/20 | `messages.52: tool_use-outside-assistant` ×2, `messages.52: dangling-tool_use-at-end` |
| after | 20/20 | none |

The `dangling-tool_use-at-end` line is the independent confirmation that re-roling was not an option.

**Live API.** A minimal synthetic request through the plugin's own headers, differing only in how the notice renders its tool call: `tool_use` → **HTTP 400** with the byte-identical message; `text` → **HTTP 200**.

**Built artifact.** The same sweep run against `lib/translate/anthropic.js` after `pnpm build` is also 20/20 clean, so the fix reaches the published bundle.

**Suite.** `pnpm test` — 115 tests, 109 pass, 0 fail, 6 skipped. Includes a regression test per commit, each written before its fix and observed failing on the exact wire defect.

## Second commit: `tool_result` blocks must lead a merged user message

The same merge has an ordering hazard, and probing the API pinned down the actual rule — a **contiguous leading run**, with order *among* the results free:

| user message | result |
|---|---|
| `[tool_result, text]` | 200 |
| `[text, tool_result]` | **400** |
| `[tool_result, text, tool_result]` | **400** |
| `[text, tool_result, tool_result]` | **400** |
| `[tool_result_B, tool_result_A]` (reversed) | 200 |

The 400 is not an ordering complaint — it surfaces as ``messages.N: `tool_use` ids were found without `tool_result` blocks immediately after``, because Anthropic answers each call against the blocks that *lead* the next message.

This is reachable rather than theoretical: a parallel tool batch arrives as one result message per call, and `toAnthropicMessages()` merges consecutive same-role messages in delivery order, so any context spliced mid-batch — a runtime-context snapshot, a settled-subagent notice — lands between two results and takes the request down.

`leadWithToolResults()` regroups the results at the front of each merged user message, preserving relative order within both groups. End-to-end through the built code, the two arms differing only in block order:

| merged user message | result |
|---|---|
| `tool_result, tool_result, text` (built output) | **200** |
| `tool_result, text, tool_result` (guard undone) | **400** |

The existing merge test encoded the old delivery order, which is a shape the API rejects whenever the result answers a call

### DIFF
--- a/src/translate/anthropic.ts
+++ b/src/translate/anthropic.ts
@@ -76,12 +76,20 @@ export function toAnthropicMessages(messages: readonly TranslatableMessage[]): A
           blocks.push({ type: 'text', text: block.text })
           break
         case 'tool-call':
-          blocks.push({
-            type: 'tool_use',
-            id: String(block.id),
-            name: block.name,
-            input: parseToolInput(block.arguments),
-          })
+          // Anthropic accepts `tool_use` only in assistant messages, and only
+          // when a matching `tool_result` follows. A tool call in any other
+          // role is replayed narrative — a settled subagent's closing message
+          // spliced into the parent as a user-role notice carries the calls it
+          // died holding, which no result will ever answer — so it rides as
+          // descriptive text instead of a call the API would reject.
+          blocks.push(role === 'assistant'
+            ? {
+                type: 'tool_use',
+                id: String(block.id),
+                name: block.name,
+                input: parseToolInput(block.arguments),
+              }
+            : { type: 'text', text: `[tool call ${block.name}: ${block.arguments}]` })
           break
         case 'tool-result':
           blocks.push({
--- a/src/translate/anthropic.ts
+++ b/src/translate/anthropic.ts
@@ -54,11 +54,35 @@ function parseToolInput(raw: string): Record<string, unknown> {
 }
 
 /**
+ * Move a user message's `tool_result` blocks into one contiguous run at the
+ * front, preserving the relative order of both groups.
+ *
+ * Anthropic answers every `tool_use` against the blocks that *lead* the next
+ * message, so a block of any other kind before or between the results reads
+ * as a call left unanswered and the request is rejected. The harness merges
+ * everything queued for one user turn into a single message, and a parallel
+ * tool batch arrives as one result message per call, so any context spliced
+ * mid-batch lands between two results. Restoring the run here keeps that
+ * independent of delivery order. Order *among* the results does not matter.
+ * @param message - one assembled user message, reordered in place.
+ */
+function leadWithToolResults(message: AnthropicMessage): void {
+  const firstOther = message.content.findIndex(block => block.type !== 'tool_result')
+  if (firstOther === -1) return
+  if (!message.content.slice(firstOther).some(block => block.type === 'tool_result')) return
+  message.content = [
+    ...message.content.filter(block => block.type === 'tool_result'),
+    ...message.content.filter(block => block.type !== 'tool_result'),
+  ]
+}
+
+/**
  * Convert harness messages into Anthropic messages. Consecutive same-role
  * messages merge into one message with multiple content blocks; tool results
- * arrive as user messages with `tool_result` blocks; system-role messages are
- * handled by {@link toAnthropicSystem} and skipped here. Reasoning blocks are
- * not replayed (v1). Images must arrive pre-resolved
+ * arrive as user messages with `tool_result` blocks, which a merged user
+ * message keeps in one leading run ({@link leadWithToolResults}); system-role
+ * messages are handled by {@link toAnthropicSystem} and skipped here.
+ * Reasoning blocks are not replayed (v1). Images must arrive pre-resolved
  * ({@link TranslatableMessage}); an unresolved ImageBlock is skipped because
  * its bytes are unreachable here.
  * @param messages - ordered conversation messages with resolved images.
@@ -118,6 +142,9 @@ export function toAnthropicMessages(messages: readonly TranslatableMessage[]): A
     const last = out[out.length - 1]
     if (last !== undefined && last.role === role) last.content.push(...blocks)
     else out.push({ role, content: blocks })
+  }
+  for (const message of out) {
+    if (message.role === 'user') leadWithToolResults(message)
   }
   return out
 }
--- a/test/translate.spec.ts
+++ b/test/translate.spec.ts
@@ -235,11 +235,13 @@ test('toAnthropicMessages: merge, tool_use input parsing, tool_result', () => {
   ])
   assert.deepEqual(messages, [
     {
+      // The merged user message leads with its tool result; the texts keep
+      // their relative order behind it.
       role: 'user',
       content: [
+        { type: 'tool_result', tool_use_id: 'call-1', content: 'result text', is_error: true },
         { type: 'text', text: 'first' },
         { type: 'text', text: 'second' },
-        { type: 'tool_result', tool_use_id: 'call-1', content: 'result text', is_error: true },
       ],
     },
     {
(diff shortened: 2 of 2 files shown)


V1ki:
Thanks @delef for the excellent report and fix! The root-cause analysis, live API verification, session replay, and focused regression coverage made this very easy to review.

I also verified the change locally: all 18 translation tests passed, including both new Anthropic regressions, and the self-contained host/client bundle completed successfully. The repository-wide `pnpm test` / `pnpm build` currently hits the same pre-existing linked DeepSeek Harness client API mismatch on `main`, unrelated to this PR.

Merged — thank you for the contribution!

PR #28, opened 2026-08-22
feat: wait out a closed rate-limit window on every route

### LINKED ISSUE #27: A closed rate-limit window fails the turn instead of waiting for the next one
## Problem

A subscription plan is rate-limit shaped by design: a 5-hour session window, a weekly window, and on some plans a per-model weekly one. Hitting a window is normal operation on a Pro/Max/Plus plan, not a failure — the window reopens at a time the provider itself discloses. Today every route treats it as a dead end and fails the turn:

```
This turn failed
claude API error (HTTP 429): {"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed your account's rate limit. Please try again later."},"request_id":"req_011CeG4D59RvAcYcSQaV2DZ9"}
RATE_LIMIT
```

The plugin already reads and renders these windows — the Subscriptions cards show used percent and reset time per window for all three providers — but nothing on the request path acts on them.

Three defects sit behind this.

### 1. No route can wait longer than a few minutes

`@deepseek-ai/dsh-llm-retry` executes the adapter-owned policy on `agent/request-error`. Its arithmetic (`lib/index.js`) is: a valid `providerRetryAfterMs` replaces local backoff **only when it is `<= maxDelayMs`**; above the cap, a `normal` policy calls `next()` — it delegates, i.e. gives up — and an `always` policy silently falls back to local backoff, so it never honours the instruction either.

What the three routes resolve to today:

| Route | Policy | Retry budget | Delay ceiling | Longest possible wait |
|---|---|---|---|---|
| `claude` | `providerRetryPolicy` override (`src/providers/claude.ts:503`), `maxRetries: 10`, 1s → 60s, jitter 0.2 | 10 | 60 s | ~5 min (1+2+4+8+16+32+60×4) |
| `codex` | no override → dsh-llm defaults | 5 | 10 s | ~16 s |
| `grok` | no override → dsh-llm defaults | 5 | 10 s | ~16 s |

A 5-hour window needs a wait measured in hours. The ceiling is measured in seconds, and any provider-supplied delay above it is not clamped — it is refused outright.

### 2. The reset time on the wire is discarded

`httpLlmError` (`src/providers/common.ts:100`) is the single classifier for all three providers. It reads exactly one signal: the `retry-after` header, and only in its numeric-seconds form (RFC 7231 also allows an HTTP-date). The body is read only as a 500-character string for the message text. Every provider-specific field naming the reset instant — Anthropic's unified-reset header, the reset seconds Codex puts in the 429 body, whatever xAI sends — is parsed by nobody, so `providerRetryAfterMs` is almost always absent and the local backoff governs.

### 3. Codex window exhaustion is classified as terminal `QUOTA`

`httpLlmError` runs `isQuotaExceededError(body)` **before** the `429` branch, so quota-shaped wording wins over the status code. Checked against the shipped classifier:

| body text | `isQuotaExceededError` | resulting code |
|---|---|---|
| `This request would exceed your account's rate limit.` | `false` | `RATE_LIMIT` |
| `{"detail":{"type":"usage_limit_reached","resets_in_seconds":9000}}` | **`true`** | `QUOTA` |
| `Weekly usage limit exceeded` | **`true`** | `QUOTA` |

`QUOTA` is not in the default retryable set, and no route adds it. So the case that most needs waiting — a ChatGPT plan's 5-hour or weekly window, which announces itself with exactly that wording — is coded as permanent and never retried at all. Claude's phrasing happens to miss the regex, which is the only reason that route reaches `RATE_LIMIT`.

The classifier is right for an API-key deployment, where an exhausted balance is terminal until someone pays. On a subscription route it is wrong by construction: there is no balance to top up, only a window with a reset time.

## Impact

A long agent run dies the moment the session window closes, mid-task, with a turn marked failed and durable history that has to be resumed by hand. The user is on a plan whose entire model is "wait for the next window", and the harness — which can hold the turn open and has the reset time in front of it — instead throws the run away. The wait is the product.

## Proposed fix

**1. One reset extractor per provider, feeding `providerRetryAfterMs`.** Each adapter passes a small parser to the shared error mapper, reading that provider's own 429 signals from the response headers and body: Anthropic's rate-limit reset header, Codex's `resets_in_seconds` / reset timestamp in the `detail` payload, xAI's `retry-after`. `retry-after` keeps working for all three and grows HTTP-date support. Exact field names get confirmed against live 429s before this lands — the parsers are per-provider precisely so a wrong guess stays contained. When a provider says nothing, nothing is invented: the policy's local backoff governs, as it does today.

**2. `429` on a subscription route is never `QUOTA`.** Check the status first: `429` classifies as `RATE_LIMIT` regardless of wording, and the quota classifier only gets a say on other statuses. A subscription plan has no terminal quota.

**3. Give each route a policy that can actually hold the wait.** New plugin config, applied to all three routes:

```yaml
- name: dsh-plugin-subscriptions
  config:
    rateLimit:
      wait: true            # default; false keeps today's seconds-scale behaviour
      maxWaitMs: 21600000   # 6 h — covers a 5-hour session window with slack
```

which resolves into each `providerRetryPolicy()` as `mode: normal`, `RATE_LIMIT` retryable, `backoff.maxDelayMs = maxWaitMs`. A reset further out than `maxWaitMs` — a weekly window days away — still fails the turn loudly rather than parking the session for a week.

The retry plugin appends `llm/retry` before waiting and `llm/retry-started` after, both carrying the scheduled delay, so the wait is already observable to renderers without new UI here.

### Trade-off worth stating

`backoff` is shared by every retryable code, so raising `maxDelayMs` also raises the ceiling of the *local* exponential backoff used for `TRANSPORT` / `SERVER` / `TIMEOUT`. On the claude route (1 s initial, 10 retries) the local sequence tops out at 512 s per attempt instead of 60 s, growing the worst case for an unrelated transient failure from ~5 min to ~17 min. Two mitigations, both cheap: keep `initialDelayMs` low and the retry budget finite, and make sure `providerRetryAfterMs` is supplied on every `RATE_LIMIT` so the provider's exact delay — used without jitter — governs the case that matters. A clean separation upstream (distinct caps for locally scheduled backoff and for accepted provider delay) would remove the trade-off entirely; worth raising there.

### Dependency

Waiting is executed by `@deepseek-ai/dsh-llm-retry`; adapters only own the policy. Without that plugin loaded nothing waits, exactly as today. The README should say so next to the config.

## Out of scope

- **Deriving the reset from the usage endpoints.** All three adapters already map `resetsAt` per window (`fetchClaudeUsage` / `fetchCodexUsage` / `fetchGrokUsage`), so a 429 with no wire signal could ask the usage endpoint instead. It costs a second network call on a failure path; revisit only if the wire signals turn out to be missing in practice.
- **A fixed fallback interval** ("retry in 5 minutes forever") — guessing at a window boundary hammers a limit that is already refusing requests.
- **A dedicated waiting indicator** beyond the `llm/retry` events the retry plugin already appends.

## Verification

- Unit tests per provider: header form, HTTP-date form, body form, absent-signal form → expected `providerRetryAfterMs`.
- A `429` carrying `usage_limit_reached` classifies `RATE_LIMIT`, not `QUOTA`.
- Resolved policy per route matches the configured `maxWaitMs`, and `wait: false` reproduces today's numbers.
- A live 429 on each route, to pin the real field names before merge.

delef:
Closes #27.

A subscription plan is rate-limit shaped by design — a 5-hour session window, a weekly one, and on some plans a per-model weekly one — so a 429 is not a dead end: the window reopens at a time the provider discloses. This makes every route wait for it.

## What was broken

**No route could hold a turn for more than a few minutes.** `dsh-llm-retry` accepts a `providerRetryAfterMs` only when it is at or below the policy's `maxDelayMs`; above it a `normal` policy delegates — it gives up rather than clamping.

| Route | Before | Longest possible wait |
|---|---|---|
| `claude` | 10 retries, 1 s → 60 s | ~5 min |
| `codex` | dsh-llm defaults | ~16 s |
| `grok` | dsh-llm defaults | ~16 s |

**The reset instant was discarded.** `httpLlmError` read one signal: `retry-after`, and only in its numeric-seconds form.

**Codex window exhaustion was classified as terminal `QUOTA`.** The quota-wording check ran ahead of the status, and `usage_limit_reached` / `Weekly usage limit exceeded` both match that classifier — which is outside the retryable set. The case that most needs waiting was never retried at all.

## What this does

**A 429 is a window, not a quota.** Status first: `429` → `RATE_LIMIT` whatever the wording. On a subscription there is no balance to top up, so the quota classifier only gets a say on other statuses (a 402 with quota wording still resolves to `QUOTA`).

**Only a signal that names the rejecting window parks a turn.** One `RateLimitResetReader` per adapter over shared parsing in the new `src/providers/rate-limit.ts`:

| Route | Read, in order | Deliberately not read |
|---|---|---|
| `claude` | `anthropic-ratelimit-unified-reset` (the subscription family Claude Code renders as "resets 3pm") → a reset named in the body | per-bucket `anthropic-ratelimit-{requests,tokens,…}-reset` |
| `codex` | `resets_in_seconds` / `resets_at` anywhere in the rejection body | `x-codex-{primary,secondary}-reset-after-seconds` |
| `grok` | `retry_after` / `resets_at` anywhere in the rejection body | `x-ratelimit-reset-{requests,tokens}` |

The right-hand column is the point. Those headers are **rollover snapshots attached to every response**, one per bucket, so on a 429 they cannot say which bucket refused. Taking the earliest picks the bucket that still had room — an `x-ratelimit-reset-requests: 0s` beside an exhausted token bucket burns the whole retry budget in seconds. Taking the latest parks a burst 429 for the hours left on an untouched primary window: `x-codex-primary-reset-after-seconds: 17000` is a 4.7-hour hold for a rejection that would have cleared immediately. Neither aggregation is right, because the information needed to choose is not in the headers. They reach the operator through the `onWarn` diagnostic instead.

**Reading is confined to a 429.** Those same headers ride a transient 500, and `dsh-llm-retry` honours `providerRetryAfterMs` for *every* retryable code — so a 529 `overloaded_error` carrying `anthropic-ratelimit-unified-reset` four hours out would sleep four hours per retry, or, once over the ceiling, give up at once. `retry-after` stays readable on any status: there it is a backoff the provider actually asked for, not a window snapshot.

**Every encoding, and by key rather than by path.** Epoch ms, epoch seconds, a delay in seconds, a Go duration (`6m0s`), an HTTP-date or RFC3339 stamp — so a provider re-encoding a field it already sends needs no code change. Body fields are found by key at any nesting depth, because these payloads move the same field between `detail`, `error` and the top level. A zero duration (`0s`) reads as "no disclosure", agreeing with the numeric path, so it cannot short-circuit the real signal behind it.

`retry-after` remains the fallback for all three and now handles its HTTP-date form. **The provider's own field wins over it**: a rejected window often carries both, and the generic header then names a short backoff that would burn the retry budget re-hitting the same closed window.

**Nothing is guessed.** A 429 that disclosed no reset this code recognizes warns through the adapter's existing `onWarn` with its rate-limit-shaped headers and the head of the body — one live hit names an unrecognized field instead of it being fixed on a guess now. The wait falls back to local backoff in that case.

**Waiting stays where it belongs.** It is provider-independent: adapters own the policy, `@deepseek-ai/dsh-llm-retry` executes it, each attempt is its own turn and the wait is already observable through the `llm/retry` events. New config, applied to all three routes:

```yaml
- name: dsh-plugin-subscriptions
  config:
    rateLimit:
      wait: true            # default
      maxWaitMs: 21600000   # 6 h — covers a 5-hour session window with slack
```

A reset further out than `maxWaitMs` — a weekly window days away — fails the turn immediately rather than parking the session; the wait is deliberately reported uncapped so the policy can refuse it outright instead of retrying into a still-closed window.

## One retry shape for all three routes

Claude Code's numbers — 10 retries, 1 s initial, 60 s cap, 20% jitter — now back every route instead of only `claude`. These are consumer subscription endpoints that shed load in bursts, and the dsh-llm defaults (5 retries, 500 ms → 10 s) give up after ~15 s, which is short for that.

| Mode | Ceiling | Delays | Total before the turn fails |
|---|---|---|---|
| `codex`/`grok` before | 10 s | 0.5, 1, 2, 4, 8 | ~15 s |
| `wait: false` | 60 s | 1, 2, 4, 8, 16, 32, 60, 60, 60, 60 | ~5 min |
| `wait: true` (default) | 6 h | 1, 2, 4, 8, 16, 32, 64, 128, 256, 512 | ~17 min |

So a 429 disclosing nothing now holds a `codex` or `grok` turn for ~17 min rather than ~15 s. `wait: false` no longer restores the previous numbers exactly — it drops back to local backoff alone, at the shared shape.

## Trade-off

`backoff` is shared by every retryable code, so raising `maxDelayMs` also raises the ceiling of *local* exponential backoff for `TRANSPORT` / `SERVER` / `TIMEOUT`: the last of the ten retries can back off 512 s instead of the 60 s cap. Bounded by the finite budget, and reached only when the provider disclosed nothing. A clean fix is upstream — separate ceilings for locally scheduled backoff and for accepted provider delay — and is worth raising there.

## Dependency

Waiting requires `@deepseek-ai/dsh-llm-retry` in the composition. Without it nothing waits and a closed window fails the turn as before; both READMEs say so next to the config.

## Verification

`pnpm build` and `pnpm test` green — 143 tests, 137 pass, 0 fail, 6 skipped (the same six as on `main`). 30 tests in the new `test/rate-limit.spec.ts` cover each provider's 429 shapes, the value/duration/date parsing, the `RATE_LIMIT`-over-`QUOTA` classification, a reset surviving a body longer than the truncated message, the diagnostic warning, and the resolved policy per route. Three of them are regressions for the fixes above:

- a 529 carrying `anthropic-ratelimit-unified-reset` four hours out classifies `SERVER` with **no** `providerRetryAfterMs`;
- a 503 with `retry-after: 30` still gets its 30 s;
- a 429 whose only signal is `x-codex-primary-reset-after-seconds` warns instead of waiting.

Still to confirm against live 429s: the exact `anthropic-ratelimit-unified-*` name, and whether Codex and xAI name a reset in the rejection body at all. Each is read by its own provider's reader, so a miss degrades that route to `retry-after` plus a diagnostic naming what the provider actually sent — it cannot mislead the others. If the warns show the body is empty on those routes, the next step is a fourth tier reading the existing usage endpoints (`fetchCodexUsage` / `fetchGrokUsage`), whose `UsageWindow.usedPercent` answers the question the snapshot headers could not: which window is actually exhausted.

https://claude.ai/code/session_01UNu8DPBvjuDQf8YBRDkgQh


V1ki:
Thanks for the rebase and the detailed follow-up on #27 — I've now reviewed this properly.

**What I verified locally:**

- Build and tests green on the PR branch as claimed (307 tests, 301 pass, 0 fail, the same 6 Keychain-gated skips as `main`), and again after a trial rebase onto today's `main` (358 tests, 352 pass, 0 fail).
- The pool interplay works exactly as you described: the route adapters still own `providerRetryPolicy` while delegating to `PoolAdapter.stream()` internally, so the widened ceiling governs pooled requests too; failover stays immediate; and the accurate `providerRetryAfterMs` directly upgrades `classifyPoolFailure`'s cooldown from the flat 5-minute guess to the real window. The uncapped wait plus `normal` mode gives the fail-fast I wanted for weekly resets.
- "Only a signal that names the rejecting window parks a turn" and confining reads to 429s are the right calls — the 529/503 regression tests pin down exactly the failure modes I'd worry about.

I'm happy with the design. Three things before merge:

**1. Rebase once more.** `main` moved since your last rebase (`parseRetryAfterMs` extraction, `OAuthEndpointError.retryAfterMs`, grok per-model effort). I did a trial rebase locally and it's small: only the first commit conflicts, in two files — `common.ts` (take your version of the retry-after hunk, but keep `parseRetryAfterMs` itself: the OAuth endpoint path now uses it) and `grok.ts` (keep both new option fields side by side). Commits 2 and 3 apply cleanly.

**2. The copilot route.** The repo grew a fourth provider since #27 was filed. `copilot.ts` calls `httpLlmError` without a reader and `CopilotAdapter` has no `providerRetryPolicy`, so it benefits from the 429→`RATE_LIMIT` reorder but gets neither the widened wait nor a reset extractor, and its pool cooldowns still fall back to the 5-minute guess. Either wire it up (even minimally: `subscriptionRetryPolicy` + `retry-after` only, with the diagnostic warn naming whatever GitHub actually sends) or state the exclusion explicitly in the PR and READMEs.

**3. A magnitude edge worth a comment.** `resetInstantFromNumber` reads any value below 1e9 as a delay in *seconds*. If a provider ever names a delay in *milliseconds* in a body field (`retry_after: 30000` meaning 30 s), it reads as 30 000 s ≈ 8.3 h — over the 6 h ceiling, so the turn fails instead of waiting 30 s. The per-provider field whitelists contain the blast radius, but a comment at the parsing site (or in the diagnostic) would help whoever debugs a wrongly refused wait later.

None of these change the design. Once it's rebased and copilot is addressed either way, I'll merge.

delef:
@V1ki done

PR #31, opened 2026-08-22
fix(codex): bound replayed tool call IDs

### LINKED ISSUE #29: Codex provider sends call_id longer than the Responses API 64-char limit
## Problem

When using a ChatGPT (Codex) subscription model, tool-call turns can fail with an HTTP 400 from the Codex backend:

```text
Invalid 'input[5].call_id': string too long. Expected a string with maximum length 64, but got a string with length 69 instead.
```

## Environment

- Plugin: `dsh-plugin-subscriptions`
- Version: `0.5.0`
- Provider: ChatGPT / Codex
- DSH profile: `web`
- Node.js: v24.19.0
- OS: Windows

## Cause

The Codex Responses API (`chatgpt.com/backend-api/codex/responses`) returns `function_call` items whose `call_id` can be **longer than 64 characters** (observed: 69). The plugin stores and re-sends that ID verbatim:

- parse side (`response.output_item.added`): `const callId = item.call_id ?? "";` — stored as the tool-call block id in conversation history
- emit side (`toResponsesInput`, `function_call` item): `call_id: String(block.id)`
- emit side (`toResponsesInput`, `function_call_output` item): `call_id: String(block.toolCallId)`

On the next request the echoed `call_id` violates the API's own 64-char limit, so the backend rejects the whole request with `string_above_max_length`.

## Suggested fix

Normalize Codex call IDs to at most 64 characters at the translation boundary, and apply the **same normalization** to both the `function_call.call_id` and the matching `function_call_output.call_id` so cross-references stay consistent. For example:

```js
const normalizedCallId = (item.call_id ?? "").slice(0, 64);
```

and use the normalized value for the tool-call block id, `function_call.call_id`, and `function_call_output.call_id` alike.

## Workaround

A local patch of the bundled `lib/index.js` with `.slice(0, 64)` at the three spots above fixes the issue (verified working).

## Notes

- The same concern applies to any other place that forwards provider-returned IDs back into request bodies (e.g. replaying `function_call` items from history).

Jstn-1g:
Closes #29.

## What changed

- Normalize tool-call IDs only when building the Codex request body, leaving the shared Responses translator (and Grok) unchanged.
- Preserve every ID that is already 64 characters or shorter.
- Replace oversized IDs with a deterministic 64-character `call_` + SHA-256 form.
- Apply one request-wide mapping to both `function_call.call_id` and matching `function_call_output.call_id` values, so their association is preserved.
- Reserve legitimate short IDs before allocating hashes and rehash on a collision.

## Why not truncate

Two distinct provider IDs can share their first 64 characters. Simple `.slice(0, 64)` would collapse them and could associate a tool result with the wrong call. Hashing at the outbound Codex boundary also repairs already-persisted histories on their next request without changing DSH's internal `CallId` or another provider's protocol.

## Tests

- `npm test` — 114 passed, 0 failed
- `npm run build` — TypeScript and production client build passed
- Regression coverage verifies short-ID preservation, the 64-character bound, call/result pairing, same-prefix long IDs remaining distinct, repeat determinism, and a forced collision with a legitimate short ID

A credentialed live Codex request was not used; the reported 69-character failure is exercised at the exact request-body boundary, and the reporter's truncation workaround already confirms that respecting the backend limit resolves the 400.

### DIFF
--- a/src/providers/codex.ts
+++ b/src/providers/codex.ts
@@ -4,7 +4,7 @@
  * endpoint.
  */
 
-import { randomUUID } from 'node:crypto'
+import { createHash, randomUUID } from 'node:crypto'
 import { attributionHeaders, EMPTY_RESPONSE_CODE, errorChain, LlmAdapter, LlmError, ReasoningEffortId } from '@deepseek-ai/dsh-llm'
 import type {
   GenerateOptions,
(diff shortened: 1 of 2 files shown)


V1ki:
Thanks @Jstn-1g for the solid fix! The deterministic hashing keeps replayed Codex tool-call IDs within the 64-character limit while preserving call/output pairing and avoiding truncation collisions. Merged — much appreciated.

PR #39, opened 2026-08-24
feat: show subscription usage in composer stats strip

kfirsch:
## Summary

Adds a compact subscription usage badge to the composer's stats line (the `conversation.composer.dock` slot), showing rate-limit window **time-remaining** and **used%** for each provider's default account.

**Before:**
```
4 turns · 45 steps | LLM 6m43s | Cache hit 94% | Input 5.6M tok · Output 23.8K tok
```

**After** (with Claude logged in):
```
4 turns · 45 steps | LLM 6m43s | Cache hit 94% | Input 5.6M tok · Output 23.8K tok | Claude 1h58m 13% · 6d18h 3%
```

## How it works

- Polls the plugin's existing `/subscriptions-auth` `status` + `usage` RPC endpoints every **15 minutes**, matching `PoolUsageTracker`'s own cache TTL so the badge causes no extra upstream traffic beyond what that cache already coalesces.
- Shows the **default account** per provider (the same account direct/non-pool routes serve) — with multiple accounts connected, the badge stays a single compact line; the full per-account breakdown already lives in Settings → Subscriptions.
- Computes time-remaining from each window's `resetsAt` timestamp: `6d18h` / `1h58m` / `42m`
- Falls back to kind labels (`5h`/`Wk`) when `resetsAt` is absent
- Returns `null` (renders nothing) when no provider has a logged-in account or usage is unsupported
- Registers with a **fresh slot id** (`subscription-usage`, order 10) — appears alongside the shipped StatsLine, never replaces it

## Changes

| File | Change |
|------|--------|
| `src/client/SubscriptionUsageBadge.tsx` | New: the badge component, with per-provider last-known-display tracking across failed polls |
| `src/client/index.ts` | Slot registration (`conversation.composer.dock`, id `subscription-usage`, order 10) |
| `src/providers/pool-usage.ts` | `PoolUsageTracker.snapshotFor()` now serves the last successful snapshot instead of throwing during a rate-limit cooldown (see review notes) |
| `test/pool-usage.spec.ts` | Tests for the stale-snapshot-served-through-cooldown behavior |

## History

This PR was originally opened against the pre-refactor single-account codebase with its own per-provider usage cache in `src/index.ts`. `main` has since landed the multi-account rewrite (`status()` → `accounts[]`, `usage(provider, account, ...)`, and the shared `PoolUsageTracker`). **Rebased and rewritten** on top of current `main`:
- The badge now targets the new API, showing each provider's default account.
- The custom per-provider cache this PR originally added is gone — superseded by `PoolUsageTracker`, which already did most of what that cache was for.
- Point 2 from the original review (a rate-limited provider's segment vanishing) turned out to still apply to `PoolUsageTracker.snapshotFor()` after the rewrite, so that's fixed there instead, with a doc comment explaining why the routing-side `quotaFor()` is deliberately left as-is (see the fix commit message for the full reasoning).

## Review notes (from @V1ki, on the pre-rebase version)

1. **Description was out of sync with the code** — the description above is now accurate against the current diff.
2. **A rate-limited provider vanishing from the badge** — fixed at the `PoolUsageTracker.snapshotFor()` level after the rebase (see above), with a new test.

Also: the out-of-scope `prepare` script commit was dropped, and the two pre-existing test-flakiness issues were filed separately as #55.

### DIFF
--- a/src/client/SubscriptionUsageBadge.tsx
+++ b/src/client/SubscriptionUsageBadge.tsx
@@ -1,0 +1,183 @@
+/**
+ * Subscription usage badge: a compact readout in the composer's stats strip
+ * (`conversation.composer.dock`), showing rate-limit window percentages for
+ * each provider's DEFAULT account (Claude, Codex, etc.). Polls the
+ * `/subscriptions-auth` `status` + `usage` endpoints on a slow interval;
+ * renders nothing when no provider has a logged-in account or usage is
+ * unsupported.
+ *
+ * Only the default account is shown — the same account direct (non-pool)
+ * routes serve — so the badge stays a single compact line even for a
+ * provider with several accounts connected. The full per-account breakdown
+ * already lives in Settings → Subscriptions.
+ *
+ * Every color resolves through a `--dsw-alias-*` design token; this component
+ * uses no locale `t` — it renders fixed short labels (provider names,
+ * "5h", "Wk") that are universal.
+ */
+import { useCallback, useEffect, useRef, useState } from 'react'
+import type { CSSProperties } from 'react'
+import type { PropsRuntime } from '@deepseek-ai/dsh-client-ui-slots'
+import type { ConnectionHandle } from '@deepseek-ai/dsh-api-remotes/client'
+import { callSubscriptionsAuth } from './SubscriptionsSection.js'
+import type { AccountStatus, ProviderStatus, ProviderUsage, SubscriptionProvider, UsageWindow } from './SubscriptionsSection.js'
+
+/** How often the badge re-reads usage; the server also shares its own cache/negative-cache across UI surfaces. */
+const POLL_INTERVAL_MS = 15 * 60_000
+
+/** Injected dependencies (slot `inject`). */
+export interface SubscriptionUsageBadgeInjected {
+  /** Connection RPC caller to reach the `/subscriptions-auth` channel. */
+  rpc: ConnectionHandle['rpc']
+}
+
+/** Props delivered by the slot outlet + inject. */
+export type SubscriptionUsageBadgeProps = PropsRuntime<'conversation.composer.dock'>
+  & Partial<SubscriptionUsageBadgeInjected>
+
+/** One provider's usage snapshot as rendered by the badge (default account only). */
+interface ProviderUsageDisplay {
+  provider: SubscriptionProvider
+  name: string
+  windows: UsageWindow[]
+}
+
+/** Brand display names (short form for the compact badge). */
+const PROVIDER_NAMES: Record<SubscriptionProvider, string> = {
+  codex: 'Codex',
+  claude: 'Claude',
+  grok: 'Grok',
+  copilot: 'Copilot',
+}
+
+/**
+ * Compact time-remaining label derived from the window's `resetsAt` timestamp:
+ * "6d18h" (days+hours), "1h58m" (hours+minutes), or "42m" (minutes only).
+ * Falls back to the scope/kind abbreviation when no reset time is known.
+ */
+function windowLabel(w: UsageWindow): string {
+  if (w.resetsAt === undefined) {
+    if (w.scope !== undefined && w.scope !== '') return w.scope
+    switch (w.kind) {
+      case 'session': return '5h'
+      case 'weekly': return 'Wk'
+      default: return 'W'
+    }
+  }
+  const ms = Math.max(0, w.resetsAt - Date.now())
+  const minutes = Math.floor(ms / 60_000)
+  const hours = Math.floor(minutes / 60)
+  const days = Math.floor(hours / 24)
+  if (days > 0) return `${days}d${hours % 24}h`
+  if (hours > 0) return `${hours}h${minutes % 60}m`
+  return `${Math.max(1, minutes)}m`
+}
+
+/** The default account of a provider's account list, when any is logged in. */
+function defaultAccountOf(status: ProviderStatus | undefined): AccountStatus | undefined {
+  if (status === undefined || status.accounts.length === 0) return undefined
+  return status.accounts.find(a => a.isDefault) ?? status.accounts[0]
+}
+
+/**
+ * The composer subscription-usage badge. Renders a compact text segment like
+ * `Claude 1h58m 13% · 6d18h 3%` (time-remaining + used%) — one segment per
+ * provider with a logged-in default account that supports usage. Returns
+ * null when no data is available.
+ */
+export function SubscriptionUsageBadge({ rpc }: SubscriptionUsageBadgeProps) {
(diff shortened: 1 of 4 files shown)


V1ki:
Thanks a lot for this contribution! The subscription usage badge is a nice addition, and the server-side cache work is solid — synchronous in-flight dedup, honoring `providerRetryAfterMs` on 429, and clearing the cache on auth changes are all well done. The tests for concurrent coalescing and rate-limit cooldown are appreciated too.

A few things I'd like addressed or discussed before merging:

**1. PR description is out of sync with the code**
- The description says the badge "polls every 60s", but the code has `POLL_INTERVAL_MS = 15 * 60_000` (15 minutes). The code is self-consistent (15-min poll matching the 15-min server cache), so please just fix the description.
- The changes table lists only 2 files and says "No changes to … the host half", but the diff touches 4 files — including the server-side usage cache in `src/index.ts`, which is the logically heaviest part of this PR. Please update the description to reflect that.

**2. A rate-limited provider vanishes from the badge**
During the 429 cooldown, `usage` throws, so `Promise.allSettled` drops that provider from `next` and `setDisplays(next)` removes its segment entirely. From the user's perspective the badge "flickers away" instead of showing the last known snapshot. Keeping the previous data for a provider whose fetch failed would be a better experience. Similarly, on the server side, when a stale `entry.value` exists it might be worth serving it instead of throwing during cooldown.

**3. Countdown labels can be up to 15 minutes stale**
`windowLabel` computes time-remaining from `Date.now()`, but the component only re-renders after each poll's `setDisplays`. So a label like `1h58m` may actually be off by up to the poll interval. Probably acceptable for a compact badge, but worth a comment or a cheaper re-render tick if you think it matters.

**4. Coalesced requests share the first caller's `AbortSignal`**
If the first caller aborts (e.g. its client disconnects), every waiter coalesced onto that in-flight promise fails together. Low probability, and the retry path recovers correctly, so this is a minor nit — but worth a note in the code.

**5. Missing test for cache invalidation on auth change**
The `authChanged` wrapper (clearing the usage cache on login/logout/account switch) is new logic with no corresponding test case. A small test asserting that a login change causes a re-fetch would round out the coverage.

Points 1 and 2 are the ones I'd like to see resolved; 3–5 are optional improvements. Thanks again!

kfirsch:
Pushed a follow-up commit (`6e31593`) addressing both required points:

**1. Description out of sync** — updated: poll interval is 15 minutes (not 60s, matching the server's own cache), and the changes table now lists all 4 files including the server-side usage cache in `src/index.ts`.

**2. Rate-limited provider vanishing** — fixed on both sides:
- Server (`usage()` in `src/index.ts`): when a cached value exists, a 429 (whether hit during an existing cooldown or on this call's own fetch attempt) now returns the stale snapshot instead of throwing. Only a provider with *no* prior successful fetch still surfaces the `RATE_LIMIT` error.
- Client (`SubscriptionUsageBadge.tsx`): keeps a per-provider last-known display across a failed poll (`lastKnownRef`), so `Promise.allSettled` rejecting for one provider no longer drops its segment. The segment is only cleared when the provider actually logs out or a fetch succeeds but reports `supported: false`/no windows.
- Added a test (`usage: a provider rate-limited after a success serves the stale snapshot instead of throwing`) covering the case: one successful fetch, cache expires, next fetch hits `RATE_LIMIT` — asserts the stale value is served rather than the call rejecting.

Ran the full suite locally (`pnpm test`) — clean aside from two pre-existing failures unrelated to this branch (reproduced identically on `main`): `device-flow.spec.js`'s whole file cascades under Node 22's test runner event-loop teardown, and one `usage.spec.js` test (`logged-out provider answers an error result`) is flaky when run alongside the rest of the suite. Happy to open a separate issue for those if useful.

Left 3–5 as optional follow-ups per your note, thanks again for the thorough review!

V1ki:
Thanks for the quick turnaround! The fixes in `6e31593a` look great — serving the stale snapshot through the cooldown on both sides (including waiters coalesced onto an in-flight request), the `lastKnownRef` handling with logout/unsupported as the only clearing signals, and the mock-timer test all address point 2 more thoroughly than I expected.

Two remaining asks:

**1. Please drop `0ceddb0c` (the `prepare` script change) from this PR.** It isn't mentioned in your follow-up comment and is out of scope for the usage badge. It may well be a legitimate fix, but `prepare` runs on every `pnpm install`, so switching it to the full `tsc && tsdown` build deserves its own PR where the motivation and install-time impact can be discussed on their own.

**2. Please open a separate issue for the two pre-existing test failures** you reproduced on `main` (the `device-flow.spec.js` cascade under Node 22's test-runner teardown, and the flaky `usage.spec.js` "logged-out provider answers an error result" case), so they don't get lost.

Once `0ceddb0c` is out, this is good to merge from my side.

kfirsch:
Both done:

1. **Dropped `0ceddb0c`** — force-pushed, PR now ends at `6e31593` again. On closer look you were right to flag it: I'd changed `prepare` to `pnpm build` because my *local* `pnpm build` (tsc + tsdown, per-file `lib/` output) didn't match what I expected from a fresh git install, but `tsdown.prepare.config.ts`'s self-contained bundle (single `lib/index.js` with no `./auth`/`./providers` sub-imports) was never actually broken — I'd compared the two builds without checking that. No functional issue there, so reverting it entirely was the right call, not just "out of scope."

2. **Opened #55** for the two pre-existing failures with full repro steps for each.

Ready for another look whenever convenient.

kfirsch:
Rebased and largely rewritten (`56831a8`, `f96a44d`) — `main` moved a lot since this was opened (the multi-account rewrite: `status()` → `accounts[]`, `usage(provider, account, ...)`, and the new `PoolUsageTracker`).

- The badge now targets the current API, showing each provider's **default account** (same one direct/non-pool routes serve).
- My original per-provider usage cache in `src/index.ts` is gone — `PoolUsageTracker` already does that job.
- Point 2 (a rate-limited provider vanishing) turned out to still apply to `PoolUsageTracker.snapshotFor()` post-rewrite, so I fixed it there instead: it now serves the last successful snapshot through a cooldown rather than throwing, both for an already-cooling-down entry and for a fetch that fails on the current call. Left `quotaFor()` (the routing/scheduling path) unchanged on purpose — explained why in the commit message, happy to discuss if you see it differently.
- New test in `test/pool-usage.spec.ts` covers a success followed by a 429, asserting the stale snapshot is served (including through repeat/forced calls during the same cooldown), plus a control case confirming a provider with no prior success still throws.

Down to 2 commits, 4 files, +271/-2 against current `main`. `pnpm build` and `pnpm test` are clean aside from the two pre-existing failures already tracked in #55 (reproduced identically on unmodified `main`, same test names/counts before and after this branch).

PR #51, opened 2026-08-29
fix: migrate to the dsh 0.1.2-alpha.1 APIs

delef:
## Why

On dsh `0.1.2-alpha.1` the published build does not load at all. The bundle carries

```js
import { …, CallId, … } from "@deepseek-ai/dsh-llm";
```

and the alpha exports `ToolCallId` instead, so this is an ESM link-time failure, not a degraded feature — the plugin's entry never evaluates and a profile that lists it fails to boot.

## What changed

- **`dsh-llm`: `CallId` → `ToolCallId`** — source and tests.
- **`dsh-host-apiproxy` is gone** (`packages/host/apiproxy` → `packages/api/gateway` + `packages/api/remotes`). `RpcResult` now comes from `dsh-client-connection` as `ConnectionRpcResult`. The Toolview components already imported from `dsh-api-remotes/client`; this finishes the move for `src/auth/rpc.ts` and the RPC specs.
- **`rpc.handle()` lost its per-channel `{ authority }` option.** The trust fence moved to the connection plugin's deployment-level `trustedHosts` config plus its Host/origin check (`packages/client/connection/src/api-request-trust.ts`), so dropping the argument preserves the loopback intent rather than widening it.
- **`dsh-client-runtime` is gone**: `ClientContext` is plain cordis `Context`, `ToolCallBlock` ships from `dsh-client-ui-conversation/client`, and the `slots` registry arrives through the `dsh-client-ui-renderer` Context augmentation — the empty-type-import idiom dsh's own client plugins use (e.g. `packages/client/ui-agent-preset`).
- **`dsh.client.inject` named `@deepseek-ai/dsh-client-runtime`**, a package the alpha no longer ships; it is replaced by `@deepseek-ai/dsh-client-ui-renderer`, which is what actually provides `slots` now. No package in the alpha injects `dsh-client-runtime` any more.
- devDependencies follow: `dsh-client-ui-renderer` added, `dsh-host-apiproxy` and `dsh-client-runtime` removed, and `host-apiproxy` leaves the bundler's `INLINE_SAFE` list.

## Compatibility

Peer ranges move to `^0.1.2-alpha.1`. One build cannot cover both releases: a range carrying a prerelease tag only matches prereleases of the same version, and the `CallId` rename would need a runtime lookup shim (`llm.ToolCallId ?? llm.CallId`) to satisfy `0.1.1-rc.2` as well. Happy to add that shim instead if you would rather keep one build working on both.

## Verification

Against a `dsh-v0.1.2-alpha.1` source build (`pnpm install && pnpm run build` in the harness checkout, its packages linked into `node_modules`):

- `tsc -p tsconfig.json --noEmit` — clean
- `tsc -p tsconfig.test.json` — clean
- `node --test lib-test/test/` — **320 passed, 0 failed, 6 skipped**
- `tsdown` — client bundle builds (96.28 kB)
- Mounted in a real `dsh --profile headless` boot on the alpha: the plugin loads and reaches its own `MISSING_CREDENTIAL` gate, where before the fix the profile failed at module link time.

https://claude.ai/code/session_019pK5Fn1JmB3Xo8RTzwQVd3

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "client": {
       "platform": "web",
       "inject": [
-        "@deepseek-ai/dsh-client-runtime",
+        "@deepseek-ai/dsh-client-ui-renderer",
         "@deepseek-ai/dsh-client-ui-settings",
         "@deepseek-ai/dsh-client-locale"
       ]
@@ -57,10 +57,10 @@
   },
   "peerDependencies": {
     "@deepseek-ai/cordis": "^4.0.1",
-    "@deepseek-ai/dsh-attachment": "^0.1.1-rc.2",
-    "@deepseek-ai/dsh-home-paths": "^0.1.1-rc.2",
-    "@deepseek-ai/dsh-llm": "^0.1.1-rc.2",
-    "@deepseek-ai/dsh-tools": "^0.1.1-rc.2",
+    "@deepseek-ai/dsh-attachment": "^0.1.2-alpha.1",
+    "@deepseek-ai/dsh-home-paths": "^0.1.2-alpha.1",
+    "@deepseek-ai/dsh-llm": "^0.1.2-alpha.1",
+    "@deepseek-ai/dsh-tools": "^0.1.2-alpha.1",
     "@deepseek-ai/schemastery": "^3.18.1"
   },
   "devDependencies": {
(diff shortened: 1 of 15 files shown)


delef:
Closing for now: dsh 0.1.2-alpha.1 is a source-only tag with no npm publish, so this migration cannot be verified against anything installable yet and it would drop 0.1.1-rc.2 support for everyone else. I will reopen (or send a fresh PR) once the alpha is published to npm. The findings stand if they are useful in the meantime: `CallId` → `ToolCallId`, `RpcResult` from `dsh-client-connection`, `rpc.handle()` without the `{ authority }` option, and `dsh.client.inject` still naming the removed `@deepseek-ai/dsh-client-runtime`.

PR #57, opened 2026-09-01
feat: add provider-backed automatic approval review

### LINKED ISSUE #56: Add opt-in provider-backed automatic review for native approval requests
## Problem

The plugin currently leaves users with two coarse choices around sandbox boundaries:

- keep `workspace-write` and answer every native escalation prompt manually; or
- run with broad access and lose the approval boundary altogether.

That is especially noisy for coding sessions where the same class of narrow, user-authorized escalation repeats. At the same time, sending every tool call to a second model would add latency and token usage even when DSH never intended to ask for approval.

Codex Guardian and Claude Code Auto Mode both use provider-owned model review to automate permission decisions, but their policies and execution models are different:

- [Codex Guardian](https://github.com/openai/codex/tree/rust-v0.151.0/codex-rs/core/src/guardian) owns its prompt construction, bounded transcript, reviewer session, and retry lifecycle.
- [Claude Code Auto Mode](https://code.claude.com/docs/en/permission-modes#auto-mode) uses a separate classifier after its permission gates. Its provider configuration includes [`environment`, `allow`, `soft_deny`, `hard_deny`, and `classifyAllShell`](https://code.claude.com/docs/en/auto-mode-config).

The shared plugin mechanism therefore should not be described as “Codex-style” or embed either provider's classifier policy. It needs one provider-neutral approval seam, with each provider owning its final review implementation.

## Desired behavior

Add an opt-in provider-backed automatic reviewer that preserves DSH's native approval mechanism:

1. Keep automatic review disabled by default.
2. Offer a persistent global `None` / `Codex` default in **Settings → Subscriptions**, retain YAML as its bootstrap/fallback, and keep the composer as a per-session override for the first implementation.
3. Capture tool metadata without calling a model, then invoke the selected reviewer only when the matching action reaches the real `approval/request` hook.
4. Route that request through a provider-neutral interface so another reviewer can be added without changing approval-hook code.
5. Let each provider own classifier policy, transcript projection, transport, retry/failure semantics, parsing, and reviewer session state.
6. Allow or reject directly only when the selected reviewer returns a conclusive decision. Missing reviewers, `ask`, and unhandled reviewer failures must fall through to the existing manual approval UI.
7. Show review progress and the final outcome in Chat without adding those audit events to model context.

This issue and its first PR publish the shared reviewer format/router and the first provider implementation, Codex. Additional providers, including Claude, will be added in separate PRs: matching each provider's behavior requires substantial source verification and reverse-engineering, provider code, fixtures, and dedicated behavioral tests.

## Why the approval boundary matters

The shared gate should not inspect every tool call. The native approval request is the existing signal that DSH has found an action requiring consent. Reusing that boundary keeps normal workspace operations model-free and avoids unnecessary latency and tokens.

The automatic reviewer also must not replace the manual provider. It is an optional first answerer in the approval waterfall; manual approval remains the generic fallback. Each provider may implement stricter decision and failure semantics internally without adding branches to the shared approval hook.

## Acceptance criteria

- [x] Default behavior is unchanged and manual (`None`).
- [x] Settings can persist a global `None` / `Codex` default across restarts; YAML remains the bootstrap before the first Settings save.
- [x] A session can select `Codex` and switch back to `None` without changing the global default.
- [x] A subagent inherits the nearest ancestor session's explicit Auto-Review choice, and an explicit child choice overrides that lineage for its descendants.
- [x] No reviewer request occurs during ordinary tool pre-execution.
- [x] Exactly one reviewer request occurs for a matching native approval request.
- [x] The shared approval gate has no Codex- or Claude-specific policy branches.
- [x] Each provider owns classifier policy, transcript projection, transport, retries, parsing, and provider-specific failure semantics.
- [x] Codex review follows the pinned upstream Guardian behavior and uses bounded context.
- [x] Review state is visible as `Auto-Review · {Provider} · Reviewing...`, followed by its result.
- [x] Review audit events are log-only and never become model-visible conversation nodes.
- [x] An unavailable or inconclusive review that is not handled by its provider falls back to the native manual prompt.
- [x] Unit tests cover routing, fallback, Codex wire shape/session reuse, denial limits, RPC state, and Chat projection.

## Out of scope

- Enabling Codex's private `free_guardian` route. The public/default `/responses` path should be used unless the backend explicitly exposes that capability.
- Automatically enabling review for existing users.
- Adding further provider implementations in the first PR; each will follow with its own source analysis, implementation, fixtures, and tests.

delef:
Closes #56.

This adds an opt-in, provider-backed automatic reviewer for DSH's native approval requests. It keeps the existing sandbox and approval boundary intact: normal tool execution remains model-free, and the selected reviewer is called only after the matching action reaches `approval/request`.

Codex Guardian is the first provider implementation. The shared mechanism is deliberately not “Codex-style”: classifier policy and lifecycle stay inside provider code so additional reviewers can be added in follow-up PRs without changing the approval hook.

## Why

Today the practical choice is either repeated manual prompts under `workspace-write`, or broad access that removes the boundary. Provider-backed review gives users a narrower middle path. Manual approval remains the path only when Auto-Review is `None`.

Reviewing every tool call would be the wrong trade-off for this plugin. Most workspace operations never need approval, so a second model call there would spend tokens and add latency without changing a decision. This implementation treats DSH's native approval hook as the routing signal. A selected reviewer is fail-closed: there is no human after that judge.

## Request flow

```text
tools/pre-execute
  └─ capture exact tool/call/arguments only (no reviewer request)

approval/request for that tool + call id
  └─ provider-neutral ApprovalReviewRouter
       └─ selected provider implementation (Codex in this PR)
            ├─ allow  → allowed-once
            ├─ deny   → rejected
            └─ unavailable / inconclusive / failure / missing callId / eviction / mismatch
                         └─ rejected (cancelled if already aborted)
                            no next() human prompt
```

The capture is consumed once and keyed by both agent and call id; a request for another tool or call cannot reuse it. The bounded recent-candidate map prevents orphaned captures from growing without limit.

## Provider-neutral boundary

`src/auto-review.ts` owns only generic orchestration:

- `ApprovalReviewer` is the provider contract;
- `ApprovalReviewRouter` resolves the session's selected reviewer;
- `AutoReviewGate` bridges `tools/pre-execute` and `approval/request`;
- provider code receives a least-privilege, read-only `ApprovalReviewAgent` view;
- only the host-side gate can append the paired audit lifecycle.

Each provider owns classifier policy, transcript projection, transport, retries, parsing, reviewer session state, and provider-specific failure semantics. The main plugin registers implementations in one list, and reviewer availability/labels flow through generic RPC state into the Settings and composer selectors. Adding another reviewer means implementing `ApprovalReviewer` in that provider and registering it; the approval hook, router, RPC validation, and client selectors do not gain provider branches.

The common decision type is limited to `allow`, `deny`, or `ask`. While a reviewer is selected, `ask`, missing correlation, capture eviction, tool-name mismatch, timeout, parse failure, and unhandled provider errors stay `rejected` (or `cancelled` if the parent request was already aborted). They do not call `next()`. Native manual approval is used only when Auto-Review is `None`.

## Codex Guardian implementation

The Codex adapter implements the first reviewer and is pinned to the public Codex CLI `rust-v0.151.0` behavior:

- model: `codex-auto-review`;
- reasoning effort: `low`;
- default ChatGPT Codex `/responses` route;
- structured `allow` / `deny` result schema with tolerant prose-wrapped JSON parsing;
- 90-second review timeout and 3 attempts;
- reusable reviewer session and prompt-cache identity per parent agent;
- only the transcript delta is sent after the first review;
- separate bounded user/assistant and tool lanes, with upstream entry/action/reason caps;
- hidden reasoning and image payloads are excluded from reviewer context;
- standard Guardian denial breaker: 3 consecutive denials or 10 of the latest 50 in one turn;
- one 401 refresh retry, then existing Codex error mapping; retryable transport/408/429/5xx/malformed-stream failures use the Guardian attempt budget.

Primary source anchors:

- [`guardian/mod.rs`](https://github.com/openai/codex/blob/rust-v0.151.0/codex-rs/core/src/guardian/mod.rs) — timeout, transcript budgets, recent-entry policy;
- [`guardian/review.rs`](https://github.com/openai/codex/blob/rust-v0.151.0/codex-rs/core/src/guardian/review.rs) — attempts and decision lifecycle;
- [`guardian/review_session.rs`](https://github.com/openai/codex/blob/rust-v0.151.0/codex-rs/core/src/guardian/review_session.rs) — reusable session behavior;
- [`guardian/prompt.rs`](https://github.com/openai/codex/blob/rust-v0.151.0/codex-rs/core/src/guardian/prompt.rs), [`policy_template.md`](https://github.com/openai/codex/blob/rust-v0.151.0/codex-rs/core/src/guardian/policy_template.md), and [`policy.md`](https://github.com/openai/codex/blob/rust-v0.151.0/codex-rs/core/src/guardian/policy.md) — prompt and output contract;
- [`truncate.rs`](https://github.com/openai/codex/blob/rust-v0.151.0/codex-rs/utils/string/src/truncate.rs) — approximate-token/UTF-8 truncation behavior.

The bundled policy snapshot is explicitly attributed to the Apache-2.0-licensed [OpenAI Codex source](https://github.com/openai/codex/blob/rust-v0.151.0/LICENSE).

### Intentional differences from Codex

- This does not claim or enable the private `free_guardian` capability. Basic review uses the normal `/responses` route, matching Codex when that gated capability is unavailable.
- DSH does not expose Codex `model_specialty` on an approval request, so the standard Guardian denial thresholds are used; cyber-specialty thresholds are not guessed.
- The transcript is projected from DSH's typed surface events into the same bounded evidence lanes instead of consuming Codex-native rollout items.

## Architecture check against Claude Code

Claude Code uses a separate classifier after its permission gates rather than Codex Guardian. Its public configuration includes `environment`, `allow`, `soft_deny`, `hard_deny`, and `classifyAllShell`: [permission modes](https://code.claude.com/docs/en/permission-modes#auto-mode), [auto mode configuration](https://code.claude.com/docs/en/auto-mode-config), and [permission evaluation order](https://code.claude.com/docs/en/permissions#how-rules-evaluate).

The current provider contract gives a future Claude implementation the exact pending action, typed session/surface access, approval reason, cancellation signal, and `allow` / `deny` / `ask` result vocabulary. The provider can own classifier prompt/configuration, transcript selection, transport, retries, and failure handling without branching the generic approval gate.

No Claude behavior is implemented in this PR. Claude and other providers will be separate follow-ups because reproducing each implementation responsibly requires substantial source verification and reverse-engineering, provider code, fixtures, and dedicated behavioral tests. The shared routing/configuration/UI path is published now together with the first Codex implementation.

## Failure and approval semantics

Automatic review is an optional first answerer. When it is selected, it is the last answerer for that request:

| Reviewer state | Approval result |
|---|---|
| `allow` | `allowed-once` |
| `deny` | `rejected` |
| no reviewer / disabled (`None`) | native manual provider |
| inconclusive / `ask` | `rejected` |
| missing `callId` / eviction (64) / tool-name mismatch | `rejected` |
| timeout / transport / malformed response | `rejected` |
| parent request already aborted | `cancelled` |

No provider error can silently grant access. A selected-reviewer failure stays denied. A human prompt appears only when Auto-Review is `None`.

## Configuration and UI

The global default lives in **Settings → Subscriptions → Default Auto-Review** and is disabled by default (`None`). The selection is persisted at `~/.dsh/plugins/subscriptions/auto-review.json` with owner-only permissions and atomic replacement, so it survives restarts.

The YAML option remains an optional bootstrap/fallback until Settings saves a value:

```yaml
- name: dsh-plugin-subscriptions
  config:
    autoReview: none   # none (default) or codex
```

The composer keeps an **Auto-Review · None/Codex** selector as a per-session override. Subagents inherit the nearest ancestor session's explicit Auto-Review choice. The same lineage behavior applies to **Speed**; Fast is still enabled only when the child session's current model advertises fast-tier support.

When a review is routed, Chat shows one log-only status node:

```text
Auto-Review · Codex · Reviewing...
```

It updates in place to `Allowed`, `Denied`, `Unavailable`, or `Cancelled`. The underlying paired audit events have no surface operation, so they are persisted for UI replay but never added to model-visible conversation context.

The client registers the projection against both supported host shapes: DSH 0.1.1-rc.2 `conversationEvents` and the 0.1.2-alpha `uiConversation.events` service.

## Other changes

- Adds typed RPC endpoints for per-session reviewer state and the persistent Settings-owned global default.
- Resolves Auto-Review and Speed through DSH's durable `session.header.parentSession` lineage, with the nearest explicit child/ancestor choice winning.
- Adds `@deepseek-ai/dsh-user-approval` as a peer/dev dependency.
- Advances `CODEX_CLIENT_VERSION` from `0.147.0` to the pinned Guardian compatibility version `0.151.0`.
- Documents Auto-Review in the English and Chinese READMEs.

## Verification

```text
pnpm build
pnpm test
```

- production TypeScript build and client bundle: pass;
- 390 tests total: 384 passed, 0 failed, 6 skipped;
- the six skips are environment-dependent Claude credential-store cases shadowed by the machine's Keychain entry;
- `git diff --check`: clean.

New coverage includes:

- default-off YAML bootstrap, persistent Settings global default, and per-session RPC overrides;
- nearest-ancestor inheritance and child overrides for Auto-Review and Speed;
- startup ordering: the first approval waits for the persisted reviewer selection;
- provider-neutral routing, dynamic reviewer discovery, and duplicate reviewer rejection;
- no review during pre-execution;
- exact tool/call matching and one-shot capture;
- allow, deny, ask, missing callId, eviction, mismatch, failure, and abort fail-closed paths; `None` still uses native manual approval;
- paired log-only activity events and Chat result formatting;
- Codex wire route, headers, metadata, model, effort, schema, and policy;
- reviewer-session reuse and transcript deltas;
- denial circuit breaker and reset behavior;
- TypeScript structural compatibility without `unknown as` casts.

## Manual smoke test

1. Link/build this branch and restart Studio/web.
2. Open **Settings → Subscriptions**, select **Default Auto-Review · Codex**, restart once, and confirm the selection persists.
3. In a session, override Auto-Review or Speed in the composer, spawn a subagent, and confirm it inherits the nearest parent choice; set an explicit child choice and confirm its descendants inherit that instead.
4. Leave the sandbox on `workspace-write` and request a narrow write outside the workspace so DSH emits a native approval request.
5. Confirm Chat shows `Reviewing...` followed by the result; a conclusive allow/deny should not also show the manual prompt.
6. Select **None** and repeat; the native manual prompt should appear unchanged.
7. Make the Codex reviewer unavailable (logged out, timeout, or parse failure) and confirm the action stays denied with `Unavailable` in Chat, not a human prompt.


delef:
### Follow-up changes since the current PR description

This comment covers only behavior added or changed after the current PR description was written.

#### Delegated subagent approval semantics

- DSH creates delegated children with approval policy `never`. The plugin now snapshots the parent session's effective Auto-Review choice at child-session creation and reopens the native approval waterfall to `ask` only when a registered machine reviewer is selected.
- `None` or an unavailable reviewer implementation preserves DSH's `never` policy as the global fail-closed fallback.
- A child reviewer `ask`, timeout, transport failure, malformed result, or other unavailable outcome no longer attempts to create a hidden manual prompt. It resolves fail-closed; an aborted request remains `cancelled`.
- The delegated system context now states that approval may be answered by the configured machine reviewer, while denied or unavailable reviews remain denied.

#### One model-visible Bash call for sandbox escalation

- After a foreground Bash call returns a structured sandbox denial, the plugin performs the sanctioned escalation retry through a nested `ctx.tools.execute()` call and projects its final result back onto the original root call. The model history and Chat therefore retain one Bash call/card instead of a denial call followed by a second model-emitted Bash call.
- The retry is still grounded in a real denial and advances only to the next sandbox rung (`read-only` → `workspace-write` → `danger-full-access`). Each further rung requires another structured denial and another approval review; no command starts speculatively with full access.
- A live delegated-session trace exposed an additional edge case: subagents may emit `sandbox_permissions: "workspace-write"` while the call also actually runs and is denied under `workspace-write`, without any widening approval. The gate now treats requested mode == actual denied mode as the current rung and performs the internal retry to `danger-full-access`. A requested mode inconsistent with the reported execution mode remains on the native path rather than being guessed.
- With Auto-Review disabled or no registered reviewer selected, the original DSH denial/retry flow is unchanged.

#### Settings ordering

- The Proxy settings section now remains above Default Auto-Review in **Settings → Subscriptions**.

#### Updated verification

- `pnpm build`: pass
- `pnpm test`: 387 total, 381 passed, 0 failed, 6 environment-dependent Claude Keychain skips
- `git diff --check`: clean
- The same-mode delegated denial is covered by the focused regression test and was verified red before the fix and green after it.

delef:
Screenshot 2026-09-02 at 1 20 39 PM
Screenshot 2026-09-02 at 1 21 41 PM

V1ki:
Thanks for the work on this. Closing as out of scope for this plugin; see the reasoning in #56. The approval gate, subagent policy changes, and in-place sandbox escalation retry belong in a dedicated plugin that can call this plugin's providers through `ctx.llm.stream()`.

A couple of pieces in this branch would be welcome as separate small PRs if you'd like to send them:

- Speed inheritance along `session.header.parentSession`;
- devDependencies moving from local `link:` paths to published `0.1.2-alpha.x` versions;
- the `CODEX_CLIENT_VERSION` bump, if 0.151 is the version the adapter should be mirroring today.

PR #63, opened 2026-09-03
fix(codex): keep session headers stable across turns

Frost2026:
Codex previously generated a new `session-id` header for every request, even within the same DSH session. This change preserves session identity across turns: existing UUIDs are retained, other non-empty session IDs map deterministically to a SHA-256-derived UUIDv8, and missing/empty IDs receive fresh random UUIDs. The request body's `prompt_cache_key` is unchanged.

Scope is limited to Codex. Grok's proposed extra headers have been removed because its Responses requests already carry `prompt_cache_key`, which xAI documents as equivalent to `x-grok-conv-id`. UUID formatting is a client convention, not a claim that gateways require it.

Validation:

- Registered the regression tests in `test/index.ts`; helper and adapter dispatch tests cover repeated/different/missing/empty IDs, UUID preservation, unchanged cache keys, and no extra Grok headers.
- `pnpm test`: 382 tests, 376 passed, 6 environment-dependent skips, no failures or cancellations. `pnpm build` passed.
- Live Codex comparison on `gpt-5.6-luna`: eight requests, all HTTP 200 and completed. After initial requests, the stable-header arm reported 2816 cached tokens on 3/3 requests; the random-header arm on 1/3. Both arms had stable body cache keys. This is a small observational comparison, not proof of a universal cache improvement or the cause of previously reported misses.
- Method, exact results, limitations, and an opt-in reproduction command are recorded in `docs/pr63-verification.md`.


delef:
@V1ki this looks important for real agentic sessions. I reproduced near-zero prompt cache hit locally on the current v0.6.0 Web runtime. The installed adapter sends a stable prompt_cache_key but generates a fresh session-id header for every request, matching this PRs diagnosis. Please prioritize review/merge if possible. I am validating the branch locally now.

PR #71, opened 2026-09-05
fix(responses): send strict: false on function tools

KyoMio:
## Problem

In the reported Codex session, the model filled in every declared property on every observed tool call, including optional fields the harness expects to be omitted.

When `strict` is omitted, OpenAI Responses attempts to normalize compatible schemas into strict mode, with a non-strict fallback for incompatible schemas. The affected harness tools express optional parameters by leaving them out of `required`, without making those fields nullable. Strict normalization can therefore change their optional behavior. See the [OpenAI documentation](https://developers.openai.com/api/docs/guides/function-calling#strict-mode).

The failure was observed with DSH 0.1.2-rc.1, whose `bash` schema includes optional `sandbox_permissions` and `justification` fields for sandbox escalation. Those two are only valid as a one-shot retry of a command the sandbox just denied. When a model emits them unconditionally:

- the first call carries `justification: ""` → `Error: invalid justification: expected a non-empty sentence`
- every later call carries `sandbox_permissions: "workspace-write"` while the session already runs under `workspace-write` → `Error: sandbox escalation to "workspace-write" is not strictly wider than this call's current "workspace-write" mode`

The command never runs. Observed with `codex/gpt-6-astra` on DSH 0.1.2-rc.1: 22 consecutive `bash` calls in a single turn, every one rejected before execution, the agent stuck re-issuing the same `pwd && git status --short`.

## Evidence

One long session, same model and account throughout; the only thing that changed between windows is this patch. Counting tool calls that filled **all** declared properties, across every tool with 3+ parameters (`bash`, `read`, `grep`, `subagent`, …):

| window | build in the profile | calls filling every property |
| --- | --- | --- |
| before | 0.7.0 from npm | 107 / 107 (100%) |
| patch live | 0.7.0 + `strict: false` | 34 / 50 (68%) |
| patch reverted | 0.7.0 from npm | 23 / 23 (100%) |

Under the patch the model omits optional parameters normally — `bash` drops to 0% full-fill and the escalation errors stop. Reverting brings both back. The effect across multiple tools supports this explanation, although the session windows are not a fully controlled experiment.

## Fix

Explicitly send `strict: false` on function tools in the Codex and Copilot Responses request builders, preserving the harness schema and its optional fields. The shared translator leaves the field absent unless a provider opts out. Grok retains its previous request shape: xAI has different strict-schema semantics and treats fields omitted from `required` as optional.

This avoids OpenAI's automatic strict normalization; it does not guarantee that a model will never voluntarily supply an inappropriate optional argument. Making all harness schemas strict-compatible would require broader schema and runtime changes upstream. No sandbox validation or approval checks are bypassed.

## Validation

- `pnpm test`: 396 passed, 6 skipped, 0 failed.
- `pnpm build`: passed.
- Added a regression test using required command fields and optional, non-nullable escalation fields. It checks schema preservation and the final request policy for Codex, Copilot, and Grok.
- Independent maintainer API comparison using the exported session's bash schema, `gpt-6-astra`, medium reasoning, and `pwd && git status --short`: all three baseline calls supplied escalation arguments; all three patched calls omitted them. One baseline call requested `danger-full-access`. The returned commands were not executed. These were fresh requests, not a replay of the entire failing session.

Related: #7 tracks the same failure class and complementary DSH core improvements. This PR addresses the adapter's strict-normalization contribution, not every possible cause of unnecessary escalation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/providers/codex.ts
+++ b/src/providers/codex.ts
@@ -652,7 +652,7 @@ export function codexRequestBody(
     // This is request-shape consistency, not a claim that Codex rejects the
     // controls when no tools are supplied.
     ...options.tools !== undefined && options.tools.length > 0
-      ? { tools: toResponsesTools(options.tools), tool_choice: 'auto', parallel_tool_calls: true }
+      ? { tools: toResponsesTools(options.tools, { strict: false }), tool_choice: 'auto', parallel_tool_calls: true }
       : {},
     ...options.reasoningEffort !== undefined
       ? { reasoning: { effort: String(options.reasoningEffort), summary: 'auto' } }
--- a/src/providers/copilot.ts
+++ b/src/providers/copilot.ts
@@ -470,7 +470,7 @@ export function copilotResponsesRequestBody(
     ...resolved.instructions !== undefined ? { instructions: resolved.instructions } : {},
     input: resolved.input,
     ...options.tools !== undefined && options.tools.length > 0
-      ? { tools: toResponsesTools(options.tools), tool_choice: 'auto' }
+      ? { tools: toResponsesTools(options.tools, { strict: false }), tool_choice: 'auto' }
       : {},
     ...options.maxTokens !== undefined ? { max_output_tokens: options.maxTokens } : {},
     // The Responses wire spells the effort nested; only advertised efforts
--- a/src/translate/responses.ts
+++ b/src/translate/responses.ts
@@ -163,6 +163,13 @@ export function toResponsesInput(
 
 /**
  * Map harness tool schemas to Responses function tools.
+ *
+ * `strict` is sent as `false` explicitly: the Responses API defaults function
+ * tools to strict mode, which demands every property be emitted on every call.
+ * Harness schemas mark optional fields by omitting them from `required` (and
+ * never as nullable), so under strict mode the model fills them all — e.g. a
+ * bash call carrying `sandbox_permissions` with no denial to justify it, which
+ * the harness rejects before anything runs.
  * @param tools - tool schemas from the request.
  * @returns Responses `tools` array entries.
  */
@@ -172,6 +179,7 @@ export function toResponsesTools(tools: readonly ToolSchema[]): Record<string, u
     name: tool.name,
     description: tool.description,
     parameters: tool.parameters,
+    strict: false,
   }))
 }
 
--- a/src/translate/responses.ts
+++ b/src/translate/responses.ts
@@ -164,22 +164,24 @@ export function toResponsesInput(
 /**
  * Map harness tool schemas to Responses function tools.
  *
- * `strict` is sent as `false` explicitly: the Responses API defaults function
- * tools to strict mode, which demands every property be emitted on every call.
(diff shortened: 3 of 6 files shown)


V1ki:
Thanks for the detailed reproduction and A/B/A comparison. I checked the implementation, the published `dsh-tool-bash@0.1.2-rc.1` schema, and the API documentation. The core premise is supported, and explicitly sending `strict: false` is a reasonable fix for the OpenAI Responses path.

The two escalation fields are optional and non-nullable in the bash schema. I also reproduced both reported validation errors using the locally installed sandbox validation functions; omitting the escalation arguments passes validation. This confirms the failure mechanism, although I have not independently replayed the reported 22-call loop or the model-side A/B/A experiment.

I tested commit `bc528e0`: **395 tests passed, 6 skipped, and the build passed**. A separate check with a representative bash schema also confirmed that the translator preserves the optional fields and the original schema while adding `strict: false`.

Before merging, I suggest three small follow-ups:

1. **Add a regression test with optional fields.** The two updated assertions currently use only `{ type: 'object' }`. Please include a bash-like schema with required command fields and optional `sandbox_permissions` / `justification`, and assert that the outgoing tool has `strict: false` while its schema and `required` list remain unchanged.
2. **Account for Grok as well as Copilot.** `toResponsesTools()` is shared by Codex, Copilot, and Grok. [xAI documents different semantics](https://docs.x.ai/developers/model-capabilities/text/structured-outputs): tool arguments always conform strictly to the schema, but fields omitted from `required` remain optional. This does not establish that adding `strict: false` breaks Grok, but please verify that its endpoint accepts the field, or scope the setting to providers whose compatibility is confirmed.
3. **Qualify the description and inline comment.** [OpenAI's documentation](https://developers.openai.com/api/docs/guides/function-calling#strict-mode) says that omitting `strict` makes Responses attempt strict normalization when possible, with a non-strict fallback for incompatible schemas. Therefore, “every property on every call” is best presented as the observed behavior in this reproduction, rather than a universal Responses guarantee. Explicit `strict: false` avoids that normalization; it does not guarantee that the model will never voluntarily supply an inappropriate optional argument.

I support the fix direction. The remaining points are regression coverage, provider compatibility, and making the explanation precise.

V1ki:
Thank you very much for finding this and for the detailed reproduction and A/B/A evidence. I consider this an important issue: it can block otherwise valid tool calls and lead to unnecessary requests for broader sandbox permissions. My own session showed the same failure pattern, and the independent API comparison supported your fix.

Given the impact, I made the review follow-ups directly on this PR and merged it: added regression coverage for optional escalation fields, scoped the strict opt-out to Codex and Copilot while preserving Grok's existing behavior, and clarified the description. The final change passed 396 tests (6 skipped) and the build.

I wanted to get this fix in promptly rather than leave the issue affecting users while waiting for another round of changes. Thank you again for identifying it, investigating it thoroughly, and contributing the fix — I really appreciate it!

PR #81, opened 2026-09-11
fix(auth): mount /subscriptions-auth ourselves so the channel survives dsh 0.1.5

### LINKED ISSUE #80: 0.8.0 on dsh 0.1.5-rc.1: every /subscriptions-auth/* RPC returns 405, login impossible
## Symptom

On dsh `0.1.5-rc.1` (npm `latest`), the Subscriptions settings page renders, but every provider
card sits at `Checking...` and the Claude card shows:

```
transport failure for /subscriptions-auth/login: HTTP 405
```

## What I measured

Clicking **Log in** produces this, captured from the browser's network panel — it is not just
`login`, it is the whole channel:

```
POST /subscriptions-auth/speed    -> 405 Method Not Allowed
POST /subscriptions-auth/status   -> 405 Method Not Allowed
POST /subscriptions-auth/proxyGet -> 405 Method Not Allowed
POST /subscriptions-auth/login    -> 405 Method Not Allowed
```

Probing the host directly agrees: `GET /subscriptions-auth/login` → 404, `POST` → 405. That is
the static handler answering, i.e. nothing is registered at that path — the node half's RPC
channel never mounts.

The server log is silent: no load error, no stack, no warning. The plugin appears in the
composed tree as `llm-subscriptions` and the client half clearly loads, since the settings tab
and all four provider cards render.

## Probable cause

`0.8.0` declares:

```json
"@deepseek-ai/dsh-llm":   "^0.1.1-rc.2 || ^0.1.2-alpha.1",
"@deepseek-ai/dsh-tools": "^0.1.1-rc.2 || ^0.1.2-alpha.1"
```

The host ships `0.1.5-rc.2` of those, which neither range admits. The node half injects only
`["llm"]` and that service is present, so it is not a missing-inject unload — it looks more like
the RPC transport contract moved between 0.1.2 and 0.1.5.

## Ruled out

- Not a duplicate-module problem from another plugin: `@leetoners/dsh-ui-subagent-monitor` drags
  in a whole `0.1.0-rc.8` family including `dsh-api-remotes` and `dsh-host-apiproxy`, but
  disabling it changes nothing — byte-identical 404/405 with it on and off.
- Not the browser-trust fence: the page is loaded through the tokenised URL and every other
  surface works.

## Question

Does `0.8.0` support dsh `0.1.5-rc.x`? If the peer ranges are simply stale, saying so would be
useful — and if the transport really did change, is a 0.9.x planned? Happy to test a branch.

## Environment

dsh `0.1.5-rc.1` (`@deepseek-ai/dsh`), `dsh-plugin-subscriptions@0.8.0`, web profile, Windows 11,
pnpm 9.15.9, node 24.

dhanjit:
Fixes #80.

## Symptom

On dsh `0.1.5-rc.1`, every `/subscriptions-auth/*` endpoint answers **405**, the Settings page
sits at `Checking...` on all four cards, and login is impossible. The server log is completely
silent. The node half loads fine — the `claude` provider still appears in the composer.

## Root cause

Not the peer ranges, not a missing service, not the `{ authority: 'loopback' }` option (0.1.5
ignores the third argument). The channel registration throws, and cordis swallows it.

`HostConnectionService.register()` — unchanged since 0.1.1 — ends in:

```js
return owner.effect(() => owner.webServer.register(route), `client-connection: ${channel} rpc channel`)
```

`owner` is `this.ctx` read through a cordis **shadow** context. `createShadow()` rebinds the
service's `ctx` property, and the context proxy's lookup then starts at
`(ctx[symbols.shadow] ?? ctx).fiber` — i.e. at the **connection plugin's own fiber**, never at
the caller's. So whether `owner.webServer` resolves depends entirely on what
`@deepseek-ai/dsh-client-connection` injects, and that changed:

| dsh-client-connection | its top-level `inject` | `owner.webServer` |
| --- | --- | --- |
| 0.1.1-rc.2 | `['webServer']` | resolves |
| 0.1.2-alpha.3 | `['webServer', 'credentials']` | resolves |
| **0.1.5-rc.2** | **`['credentials']`** — web transport moved into a nested `ctx.inject(['webServer'], …)` child fiber | **throws** |

The throw is `cannot get property "webServer" without inject`, raised inside `Fiber._execute`
while the plugin fiber is running, which is why nothing reaches the log. With no route
registered, POSTs fall through to the static file handler in `@deepseek-ai/dsh-client-modules`
(`if (method !== 'GET' && method !== 'HEAD') return { status: 405 }`) — the actual source of the
405. Nothing in the connection layer or the web server emits 405 at all.

Captured on the live host, before the fix:

```
[PROBE] handle THREW: Error: cannot get property "webServer" without inject
    at Fiber.<anonymous> (…/@deepseek-ai/dsh-client-connection/lib/index.js:618:35)
    at Proxy.register (…/@deepseek-ai/dsh-client-connection/lib/index.js:618:16)
    at Object.handle (…/@deepseek-ai/dsh-client-connection/lib/index.js:543:39)
    at Fiber.<anonymous> (…/dsh-plugin-subscriptions/lib/index.js:1460:101)
```

Worth stating plainly: **adding `webServer` to this plugin's inject list does not fix it.** I
tried that first. The nested inject callback then gets a working `ctx.webServer`, but
`owner.webServer` inside `register()` still throws, because the shadow context is not ours.
`rpc.handle()` simply cannot mount a prefix channel on 0.1.5 for any caller — which is probably
why nothing in the host uses it; `dsh-api-gateway` uses `rpc.intercept('/api', …)`, and that path
never touches `owner.webServer`.

## The fix

Register the same prefix route on `webServer` directly. Everything the browser sees is unchanged:
same `/subscriptions-auth` path, same `client-request` / `server-response` envelope, same
`rpcId` echo, and the same fence — `connection.requestRejection()` is the exact method the host's
own `/api` route calls, so Host/Origin checks and browser auth still run in front of the channel.

* `webServer` joins `connection` in the nested inject, because the channel is an HTTP route: with
  no web server there is nothing to mount on. Headless compositions have neither service, so they
  are unaffected.
* Hosts predating `requestRejection` (0.1.1-rc.2) keep the `rpc.handle` path, which still works
  there.
* The buffered body is capped at 1 MiB. Every request on this channel is small — ids, an account
  key, a pasted callback URL, a proxy config; image and video bytes travel out in the response,
  never in.

## Verification

Built from this branch, packed, installed into a throwaway profile
(`@deepseek-ai/dsh-base` + `@deepseek-ai/dsh-web-app` + this plugin) on dsh `0.1.5-rc.1`:

| request | before | after |
| --- | --- | --- |
| `POST /subscriptions-auth/status` (unauthenticated) | `405` | `401 unauthorized` (the fence) |
| `POST /subscriptions-auth/login`, `/speed`, `/proxyGet` | `405` | `401` |
| `GET /subscriptions-auth/speed` | `404` | `404` |

With the browser session cookie the page itself uses:

```
$ curl -b cookies.txt -X POST -H 'content-type: application/json' \
    -d '{"type":"client-request","rpcId":"probe-3","method":"proxyGet","payload":{}}' \
    http://127.0.0.1:3090/subscriptions-auth/proxyGet
{"type":"server-response","rpcId":"probe-3","result":{"ok":true,"value":{"enabled":false,"url":"","passwordSet":false,"bypass":[]}}}
```

…and the dispatch guards still answer as business failures rather than transport failures:

```
{"result":{"ok":false,"error":{"code":"bad-request","message":"unknown /subscriptions-auth endpoint \"notARealEndpoint\"",…}}}
{"result":{"ok":false,"error":{"code":"bad-request","message":"method \"login\" does not match endpoint \"speed\"",…}}}
```

I did not complete an OAuth login — proving the channel mounts and dispatches was the goal, and no
credential was touched.

`pnpm test` is green: **425 tests, 0 failures**, including 6 new ones in
`test/channel-route.spec.ts` that drive the registered route over a real `node:http` server
(envelope round trip, `rpcId` echo, method/path disagreement, the fence answering 401, non-POST
and non-endpoint paths, and the 400/415 body guards).

## Notes for the reviewer

* **One new devDependency**, `@deepseek-ai/dsh-host-webserver@0.1.2-alpha.3`, pinned like the
  other dsh devDeps. It is type-only — same role `@deepseek-ai/dsh-client-connection` already
  plays — and it is what lets the compiler check the route shape against the host instead of
  against a hand-written guess, which is precisely the kind of drift that caused this bug. Happy
  to drop it for a locally declared interface if you would rather not add it; the lockfile delta
  is purely additive (124 added, 8 re-keyed peer hashes).
* **The declared peer ranges are untouched.** `^0.1.1-rc.2 || ^0.1.2-alpha.1` still does not admit
  the 0.1.5 the host ships, so they remain misleading — but they are not what breaks the channel,
  pnpm only warns, and widening them is a separate call that belongs in its own change.
* This does not fix `rpc.handle()` for other plugins. Any plugin registering a prefix channel on
  0.1.5 hits the same wall; it likely deserves an upstream issue against the harness.


V1ki:
Thanks for the detailed root-cause analysis and the proposed fix. This is now covered by an independent implementation in e8d9efa7adfe57200d09fe79d1f941496861bb8b (2026-09-12), included in the v0.9.0 tag and retained in current main.

Both this PR and #82 branch from 1a2777c (v0.8.0), before that change. Rather than mounting the legacy /subscriptions-auth prefix directly on webServer, main registers each endpoint as an exact POST Fetch route under /api, e.g. /api/subscriptions-auth.status. All client callers, including image/video toolviews, now use the shared /api caller. This removes the connection.rpc.handle() path responsible for #77/#80.

Closing as superseded by the independent implementation, not merging this branch. Related alternative: #82; related reports: #77 and #80. Thank you for identifying the shadow-context/webServer injection failure.

Verification scope: production typecheck passes; the current local full suite passes with 436 passed / 6 skipped / 0 failed after a test-only Antigravity fixture adaptation (not yet committed). A fresh authenticated browser/OAuth end-to-end run has not been performed in this review. Compatibility with the older DSH 0.1.1-rc.2 host is a separate remaining consideration: main does not retain the legacy fallback proposed here.

Issue #53, opened 2026-08-31
support dsh-v0.1.2-alpha.2


Adol1111:
目前dsh已经把0.1.2-alpha.2上传至npm，希望尽快更新

dhanjit:
The whole alpha package set is on npm now, not just `@deepseek-ai/dsh`, and the current alpha tag is **0.1.2-alpha.3** (alpha.1 in the title is two behind).

Verified against the registry today:

| package | `alpha` dist-tag |
|---|---|
| `@deepseek-ai/dsh-llm` | 0.1.2-alpha.3 |
| `@deepseek-ai/dsh-tools` | 0.1.2-alpha.3 |
| `@deepseek-ai/dsh-attachment` | 0.1.2-alpha.3 |
| `@deepseek-ai/dsh-home-paths` | 0.1.2-alpha.3 |
| `@deepseek-ai/dsh-client-connection` | 0.1.2-alpha.3 |
| `@deepseek-ai/dsh-api-remotes` | 0.1.2-alpha.3 |
| `@deepseek-ai/dsh-client-ui-conversation` | 0.1.2-alpha.3 |
| `@deepseek-ai/dsh-client-ui-renderer` | 0.1.2-alpha.3 |

`npm pack @deepseek-ai/dsh-llm@alpha` confirms the rename is real in the published build — its export list contains `ToolCallId` and no `CallId`:

```
export { APP_IDENTITY, BlockAssembler, ..., MessageId, ProviderRequestId,
         QUOTA_EXCEEDED_CODE, ReasoningEffortId, RetryPolicySchema,
         ToolCallId, assertUsableApiKey, ... }
```

One data point for the `dsh.client.inject` item in #51: `@deepseek-ai/dsh-client-runtime` has **no** `alpha` dist-tag (only `latest` 0.0.1-rc.1 / `next` 0.1.1-rc.2), consistent with it being removed in the alpha line rather than renamed.

So the blocker stated when #51 was closed — "source-only tag with no npm publish, cannot be verified against anything installable" — no longer applies; every package that migration touches is installable via the `alpha` tag.

Context for why this bites now: DSH Desktop 2.0.4 bundles the 0.1.2-alpha runtime, so installing 0.5.3 into one of its profiles fails at ESM link time and takes the entire profile boot down with it:

```
failed to import loader entry llm-subscriptions (dsh-plugin-subscriptions):
[...]

PR #82, opened 2026-09-11
fix(transport): serve settings endpoints as an /api Fetch route on dsh 0.1.5+

Sunyata-Anatta:
## Problem

On DSH v0.1.5 the Settings → Subscriptions panel is dead: every
`/subscriptions-auth` POST comes back **405**, so no provider can be
inspected or logged in.

The cause is on the host side. `connection.rpc.handle()` mounts a channel
through `owner.webServer.register(route)`, where `owner` is the connection
plugin's own context — and since `5dda764e` (v0.1.5-alpha.1) that context
narrowed `inject` from `['webServer', 'credentials']` to `['credentials']`.
Registration throws, no route is mounted, and the POST falls through to the
SPA handler, which answers 405. This affects every consumer of
`rpc.handle`, not just this plugin. I'm reporting it to `deepseek-harness`
separately; this PR is only the plugin-side fix so the panel works on the
current host.

## Change

Exact Fetch routes take a different path into the host: they land in the
connection's own route map and are dispatched under the already-mounted
`/api` prefix, so they never touch the broken `webServer` injection. That
registry has existed since v0.1.2-alpha.1.

The node half now registers a single exact route,
`/api/subscriptions-auth`, when `connection.fetch` exists, and keeps the
legacy channel only for v0.1.1-rc.2. One route with the endpoint in the
payload, rather than one path per endpoint, keeps a single registration
instead of a path list that has to track all 18 endpoints.

The route speaks the same `client-request` / `server-response` envelope as
the host's channel bridge, carrying `{ endpoint, payload }`, so the browser
keeps calling `rpc.call` and every caller sees the unchanged result shape.
Calls aimed at the legacy channel are rewritten onto the route unless the
client handle still has rc.2's `.api` face.

Both transports share one `dispatch` closure, so there is no second copy of
the endpoint logic.

## Compatibility

| Host line | `connection.fetch` | Transport used |
|---|---|---|
| v0.1.1-rc.2 | absent | legacy `/subscriptions-auth` RPC channel |
| v0.1.2-alpha.1 … v0.1.2-rc.1 | present | `/api/subscriptions-auth` Fetch route |
| v0.1.3-alpha.1 … v0.1.3-alpha.2 | present | `/api/subscriptions-auth` Fetch route |
| v0.1.5-alpha.1 and later | present | `/api/subscriptions-auth` Fetch route (`rpc.handle` is broken here) |

`peerDependencies` is unchanged; one build still serves every line, in
keeping with #53 and `src/compat.ts`.

Two fields of the route literal need a local compat type, next to the
existing `RpcHandleCompat`:

- `requestBody` arrived in v0.1.3-alpha.1.
- `ConnectionFetchMethod` is `'GET' | 'HEAD'` until v0.1.3-alpha.2 widens it
  to include `'POST'`.

The v0.1.2-alpha types this package builds against predate both. That
runtime does serve the route regardless: `assertFetchRoute` validates only
the path, the method-list arity and duplicates, and
`createSharedFetchHandler` dispatches on
`route.methods.has(request.method)`. I widened `FetchRouteCompat` instead of
raising the devDependency, so the lowest supported line stays the type
floor and a future incompatibility still shows up in `tsc`.

## Verification

- `pnpm test`: 428 passing, 0 failing — 419 on `main`, plus 9 new tests
  across `test/rpc-fetch.spec.ts`, `test/subscriptions-rpc.spec.ts` and
  `test/rpc.spec.ts`.
- `npx tsc --noEmit` and `pnpm build` clean.
- Against **every published `@deepseek-ai/dsh-client-connection` in the
  supported range** — `0.1.1-rc.2`, `0.1.2-alpha.2`, `-alpha.3`, `-alpha.5`,
  `0.1.2-rc.1`, `0.1.3-alpha.2`, `0.1.5-alpha.1`, `-alpha.2`, `0.1.5-rc.1`,
  `0.1.5-rc.2` — a real `HostConnectionService` was constructed,
  `registerAuthRpc` was run against it, and a real POST was driven through the
  transport it selected. All ten return the `server-response` envelope with
  `ok: true` and the expected value, and all ten answer 404 for an undeclared
  method on the same path, so the registration does not over-capture.
  `0.1.1-rc.2` takes the legacy `/subscriptions-auth` channel, as intended;
  every later version takes the Fetch route with `methods: ['POST']`.
- Manually in a browser on DSH v0.1.5: Settings → Subscriptions loads and all
  four providers report status; the 405 is gone.
- Manually in a browser on DSH v0.1.2-rc.1: Settings → Subscriptions behaves
  as before. This is the narrow end of the range — no `requestBody` field, and
  `ConnectionFetchMethod` still `'GET' | 'HEAD'` — so it exercises the
  compat type against the host that declares the least.
- The v0.1.3 line takes the same Fetch route rather than falling back, so its
  behaviour did change. Every function this change touches —
  `createSharedFetchHandler`, `registerFetchRoute`, `assertFetchRoute`,
  `bridge` and `requestBodyMode` — is byte-identical between
  `0.1.3-alpha.2` and `0.1.5-rc.2`. The only difference between those two
  releases is the `inject` narrowing and the resulting move of the `/api`
  mount, which is what broke `rpc.handle` and does not touch the Fetch-route
  path. The browser check on v0.1.5 therefore covers v0.1.3's behaviour for
  this route as well.

## Out of scope

The host regression itself. It needs a fix in `deepseek-harness`
(`HostConnectionService.register` resolving `webServer` from a context that
no longer injects it), reported separately.


V1ki:
Thanks for the implementation and the extensive transport compatibility analysis. The DSH 0.1.5 authentication 405 problem is now covered by an independent implementation in e8d9efa7adfe57200d09fe79d1f941496861bb8b (2026-09-12), included in the v0.9.0 tag and retained in current main.

This PR and #81 branch from 1a2777c (v0.8.0), before that change. Main uses the same architectural direction as this PR—exact Fetch routes under the existing /api mount, avoiding connection.rpc.handle()—but registers /api/subscriptions-auth.<endpoint> per endpoint rather than one /api/subscriptions-auth route with an endpoint wrapper. The shared client caller and image/video toolviews were migrated together.

Closing as superseded by the independent implementation, not merging this branch. Related reports #77/#80 are being closed against that existing fix; #81 is the alternative direct-webServer implementation.

Verification scope: production typecheck passes; the current local full suite passes with 436 passed / 6 skipped / 0 failed after a test-only Antigravity fixture adaptation (not yet committed). This review has not performed a fresh authenticated browser/OAuth end-to-end run. Your legacy fallback for DSH 0.1.1-rc.2 remains a useful separate compatibility consideration: current main requires connection.fetch and does not preserve that fallback.

PR #99, opened 2026-09-21
fix(antigravity): preserve valid tool-call history

### LINKED ISSUE #97: [Bug][Antigravity] Tool-call failures in conversation history
## Context

This issue collects failures observed in sessions using the Google Antigravity provider with tool calls. The goal is to **discuss and confirm the expected behavior and agree on the appropriate approach** before opening a small, focused pull request.

[PR #96](https://github.com/V1ki/dsh-plugin-subscriptions/pull/96) contains experimental changes used to demonstrate and test the reported scenarios. It is **not mergeable** and should not be treated as the final proposal. I have not yet had time to prepare a pull request dedicated exclusively to these bugs.

### 1. Tool result rejected by the API

In some conversation histories, the API returns HTTP 400 when a tool result contains JSON with a top-level array or primitive value:

```text
Invalid JSON payload received. Unknown name "response" at request.contents[54].parts[0].function_response: Proto field is not repeating, cannot start list.
```

This suggests a mismatch between the tool-result format serialized by the adapter and the format accepted by the Gemini/Antigravity endpoint.

Reference: [FunctionResponse](https://ai.google.dev/api/generate-content).

### 2. Missing thought signature on replayed function calls

In turns following a tool call, the API may reject the replayed call with HTTP 400:

```text
Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:run_code`, position 82.
```

We need to determine when a `thoughtSignature` must accompany a replayed `functionCall` and how the adapter should preserve this data in conversation history.

Reference: [Gemini thought signatures](https://ai.google.dev/gemini-api/docs/thought-signatures).

## Discussion points

- What is the canonical way to serialize tool results for the Antigravity endpoint?
- Which models and tool-call patterns require `thoughtSignature` during replay?
- How should this metadata be preserved without affecting sessions, models, or providers that do not use it?
- Which regression scenarios should be covered by the focused pull request?

Reports including the model used, a minimal message/tool sequence, and sanitized payloads are welcome.

victorwads:
## Summary

Fixes two Antigravity/Gemini tool-history serialization failures:

- Normalizes every `functionResponse.response` value to a JSON object, as required by the Gemini API. Arrays, primitives, and `null` are wrapped as `{ result: value }`; non-JSON text remains `{ output: text }`.
- Preserves real replayed thought signatures and supplies Gemini's documented `skip_thought_signature_validator` sentinel for the first unsigned function call in each Gemini 3 tool-call step. This permits users to switch an existing conversation from another provider (for example GPT/Codex) to Gemini 3 without a 400 validation failure.

## Context

When serialized tool results have a top-level array, Antigravity rejects the request with an error such as:

```text
Invalid JSON payload received: Proto field is not repeating, cannot start list.
```

When a prior provider's tool call is replayed without a Gemini thought signature, Gemini 3 rejects it with:

```text
Function call is missing a thought_signature in functionCall parts.
```

The fallback sentinel is the documented approach for manually constructed or cross-model histories that do not carry Gemini signatures.

## Tests

- Covers object, array, `null`, and numeric tool-result payloads.
- Covers parallel and sequential foreign tool-call steps replayed by Gemini 3.
- Verifies that Gemini 2.5 does not receive the Gemini 3 fallback.

Closes #97

References:
- https://ai.google.dev/api/generate-content
- https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures


victorwads:
@V1ki i'm using this branch for some days, so, i think maybe it should be merged. tested A LOT on several antigravity sessions.

Most tests changing from Codex to Antigravity to Codex to Antigravity mid session.
It Works!

V1ki:
感谢 @victorwads 的贡献，也感谢在 #97 里把两个问题梳理清楚！

我先合并了 #100（它修的是同一个 `functionResponse.response` 非对象的问题，测试覆盖更全），然后把这个分支 rebase 到了最新 main（4278b1d）：

- `toolResultValue` 采用 #100 的 `{ output: value }` 封装，与现有非 JSON 路径保持一致，同时保留了你把返回类型收紧为 `Record<string, unknown>` 的改动；
- thought signature 部分原样保留。我对照 Google 文档核实过，`skip_thought_signature_validator` 是官方占位签名，"每个 step 只校验第一个 functionCall、仅 Gemini 3 严格校验"的规则也和实现一致；
- 原有测试里 `result` 键的断言改成了 `output`，其余不变。干净构建下 501 个测试全部通过。

方便的话请再看一眼 rebase 后的分支，没问题我就合并。

---

Thanks @victorwads for the contribution, and for laying out both failures so clearly in #97!

I merged #100 first (it fixes the same non-object `functionResponse.response` issue with broader test coverage), then rebased this branch onto the latest main (4278b1d):

- `toolResultValue` now uses #100's `{ output: value }` envelope, consistent with the existing non-JSON path, while keeping your tightened `Record<string, unknown>` return type;
- the thought-signature part is unchanged. I verified against Google's docs that `skip_thought_signature_validator` is the official placeholder and that the "first functionCall per step, strict on Gemini 3 only" rule matches the implementation;
- the `result` key assertions in the existing test were switched to `output`; nothing else changed. All 501 tests pass on a clean build.

Please take a quick look at the rebased branch when you have a moment. If it looks good to you, I'll merge it.

victorwads:
Thanks

